PR #10, opened 2018-05-17
FF-1086 from tibanna

carlvitzthum:
Pretty major overhaul. Lots of organizational changes and re-writes of metadata helper functions in ff_utils.py. Added tibanna_utils.py.

### DIFF
--- a/dcicutils/s3_utils.py
+++ b/dcicutils/s3_utils.py
@@ -11,7 +11,7 @@ import logging
 ###########################
 # Config
 ###########################
-s3 = boto3.client('s3', region='us-east-1')
+s3 = boto3.client('s3', region_name='us-east-1')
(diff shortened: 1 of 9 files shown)


dcicutils/s3_utils.py
```diff
@@ -11,7 +11,7 @@
 ###########################
 # Config
 ###########################
-s3 = boto3.client('s3')
+s3 = boto3.client('s3', region_name='us-east-1')
```
j1z0: Alex had a branch, that for whatever reason didn't get merged in, but moved the s3 here to s3Utils.s3.  I think that needs to happen, maybe just implement it in the branch since Alex's branch has diverged.

carlvitzthum: Done

PR #28, opened 2018-10-03
Shipping logs to ES

carlvitzthum:
Used the pre-existing Elasticsearch utils combined with some logging/structlog magic to make logs automatically ship to ES when `set_logging` is called with an `es_server` and `in_prod`

### DIFF
--- a/dcicutils/log_utils.py
+++ b/dcicutils/log_utils.py
@@ -179,6 +179,10 @@ def set_logging(es_server=None, in_prod=False, level=logging.INFO, log_name=None
         cache_logger_on_first_use=True,
     )
 
+    if log_name is None:
+        log_name = __name__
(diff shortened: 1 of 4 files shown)


Chadius:
My only comment is that the unit tests seem to rely on an external service, ES. If ES fails for any reason your tests will fail. But we're in bigger trouble if ES is down, anyway.
Looks good to me.

carlvitzthum:
> My only comment is that the unit tests seem to rely on an external service, ES. If ES fails for any reason your tests will fail. But we're in bigger trouble if ES is down, anyway.
> Looks good to me.

Yeah, it was a conscious decision to have integrated tests that depend on ES and Fourfront servers being up. These are important given the nature of the utilities repo. They are paired with non-integrated unit tests where possible

PR #56, opened 2020-01-13
New ES compatibility

willronchetti:
- Hook in ES6 compatible library


netsettler:
Has this been tested in cgap and fourfront so we know that it's a compatible change?

willronchetti:
Yes - I have these PR's open that both use this version and work with ES6. 

- https://github.com/dbmi-bgm/cgap-portal/pull/204 (builds passed with es5 and pass locally with es6)
- https://github.com/4dn-dcic/fourfront/pull/1378 (builds are broken with es5 but pass locally with es6)

netsettler:
OK. I realize a change of this kind is complex and leans heavily on whether we have adequate test coverage to know if it's a "compatible change", so I suppose if our testing is deficient somewhere, we could be left chasing and fixing bugs due to compatibility issues, but that's just inevitable.

netsettler (approved):

PR #57, opened 2020-02-10
Update s3_utils.py

SooLee:


SooLee:
Added your examples below in the comments. Thanks!


> 
> ```
> first_filename = file_list[0]
> if first_filename.endswith('/'):
> ...
> ```

I didn't do this because we would need this variable only once. 

> 
> Did some spec change or was this never true in the first place? A comment explaining what is expected here seems like it would help. namelist on a ZipFile is supposed to return a list of files. I'm guessing you sometimes see things like:
> 
> ```
> ["foo/", "file1", "file2", "file3"]
> ```
> 
> and other times thinks like
> 
> ```
> ["file1", "file2", "file3"]
> ```
> 
> and only in the former case do you want to pop the file dir out and save it for later use somehow?

Yes, previously all the zip files had a directory as the first element. Now we have a case where there is no directory in the zip file.

SooLee:
It's failing test with these errors:
```Error updating JH session: Bad status code for GET request for http://fourfront-mastertest.9wzadzju3p.us-east-1.elasticbeanstalk.com/c86c8c87-6438-4233-b4a6-6dd9bd9dfe89: 404. Reason: {'status': 'error', 'title': 'Not Found', 'code': 404, 'description': 'The resource could not be found.', '@type': ['HTTPNotFound', 'Error'], 'detail': '/c86c8c87-6438-4233-b4a6-6dd9bd9dfe89'}```

PR #58, opened 2020-02-13
C4-24 Give ff_utils a lucene API

willronchetti:
- Implements `search_es_metadata` allowing one to run a lucene query on the given `ff_env` by accessing ES directly
- NOTE: For repeat usage, see `SearchESMetadataHandler`

### DIFF
--- a/dcicutils/ff_utils.py
+++ b/dcicutils/ff_utils.py
@@ -891,23 +891,27 @@ class SearchESMetadataHandler(object):
     """
 
     def __init__(self, key=None, ff_env=None):
-        self.health = get_health_page(key, ff_env)
+        self.health = get_health_page(key, ff_env)  # expensive - do not call repeatedly!
         self.es_url = self.health['elasticsearch']
         self.client = es_utils.create_es_client(self.es_url)
 
-    def execute_search(self, index, query):
+    def execute_search(self, index, query, is_generator=False, page_size=200):
         """
         Executes lucene query on this client's index.
 
         :arg index: index to search under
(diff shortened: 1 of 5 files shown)


dcicutils/ff_utils.py
```diff
@@ -879,6 +879,40 @@ def get_health_page(key=None, ff_env=None):
...
+
+
+def search_es_metadata(index, query, key=None, ff_env=None):
+    """
+        Executes a lucene search query on on the ES Instance for this
+        environment.
+
+        :arg index: index to search under
+        :arg query: dictionary of query
+        :arg key: optional, 2-tuple authentication key
```
netsettler: I assume the 2-tuple is `(name, password)`. It's probably good to say, or to hyperlink to a description of what's allowed.

test/conftest.py
```diff
@@ -27,6 +27,8 @@ def integrated_ff():
     integrated['ff_key'] = s3.get_access_keys()
     integrated['higlass_key'] = s3.get_higlass_key()
     integrated['ff_env'] = INTEGRATED_ENV
+    # XXX: Read from config? Somewhere else?
```
netsettler: Either a config or an environment variable (defaulting to something safe like localhost but letting you redirect), though it's best if the environment variable stuff is resolved in a config file (even if in active code like a settings.py file rather than declaratively in a `.ini` file or some such.) It's good to be able to read off all these values in one place, kind of like a style sheet (CSS) rather than litter them through the code.

PR #59, opened 2020-02-20
Relax most requirements for easier use as a library in other apps.

netsettler:
To be an included library, `dcicutils` needs not to impose unnecessary requirements on its consumers. This tries to relax the ones that don't matter.

Note that this change does create a constraint to use Python 3.4 - 3.6, locking out 3.7 and beyond. I regard that as a bug fix, not an added restriction. It's not clear this would have worked for higher versions. We'll need to recheck things more carefully before we go to 3.7.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "dcicutils"
-version = "0.9.2"
+version = "0.9.3b0"
 description = "Utility package for interacting with the 4DN Data Portal and other 4DN resources"
 authors = ["William Ronchetti <william_ronchetti@hms.harvard.edu>"]
 license = "MIT"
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -12,23 +12,24 @@ packages = [
...
-python = "^3.4"
-boto3 = "1.10.46"
-botocore = "1.13.46"
-elasticsearch = "5.5.3"
-aws-requests-auth = "0.4.2"
+python = ">=3.4,<3.7"
+boto3 = "^1.10.46"
+botocore = "^1.13.46"
+# TODO: elasticsearch is on version 7. -kmp 15-Feb-2020
+elasticsearch = "^5.5.3"
```
willronchetti: This may have to stay pinned at `5.5.3` for now but it's probably okay to use caret here since it will stay below `6.0.0` where changes will be breaking.

netsettler: Yeah, I figured there could be small bug fixes that snuck in. But the caret is trying to avoid breaking changes for things that respect that. Some of these libraries turn out not to. But I bet they are careful with ES.

PR #61, opened 2020-02-26
modified get_associated_qc_metrics to include more information

Lmercadom:


dcicutils/ff_utils.py
```diff
@@ -457,65 +457,150 @@ def faceted_search(key=None, ff_env=None, item_type=None, **kwargs):
...
+    Args:
+        ignored_fields: flag to ignore 4DN custom fields from the qc metric object
+
+    Returns:
+        a dictionary of dictionaries containing the qc_metric information
+    """
+    qc_metrics = {}
+
+    if ignored_fields:
+        ignored_qc_fields = ['contributing_labs', 'schema_version', 'external_references', '@context', 'aliases',
```
netsettler: This list should be separated out into a constant at file toplevel so that it can be inspected by people wondering what's in it. You might also want to make it be a tuple in that case, just so no one is tempted to modify it. But let's assume you've done:
```
IGNORABLE_QC_FIELDS = ('contributing_labs', ...)
```
Then here you might want to do something like this (untested, just offered as a sketch):
```
if isinstance(ignored_fields, (list, tuple, set)):
    ignored_qc_fields = ignored_fields
elif isinstance(ignored_fields, bool):
    ignored_qc_fields = IGNORABLE_QC_FIELDS if ignored_fields else []
else:
    raise ValueError("Expected ignored_fields to be True, False, or a collection of field names: %s"
                     % ignored_fields)
# If we're going to a lot of membership tests, it will be faster as a set
ignored_fields.= set(ignored_fields).
```

dcicutils/ff_utils.py
```diff
@@ -457,65 +457,150 @@ def faceted_search(key=None, ff_env=None, item_type=None, **kwargs):
...
-    Given a uuid of an experiment set, return a dictionary of uuid : item
-    mappings of quality metric items.
+    Utility function to grap all the qc metrics from associated types of file such as:
+    'proccessed_files', 'other_processed_files', 'files'
+    inputs:
+        data: the metadata of a ExperimentSet or Experiment
+        associated_files: a list of the types of the files fields the qc metrics will be extracted from:
+            examples are = ['files', 'processed_files', 'other_processed_files']
+    Args:
+        ignored_fields: flag to ignore 4DN custom fields from the qc metric object
```
netsettler: If it's just a boolean, I might rename this argument to `ignore_typical_fields` or some other kind of thing that suggest you're not supposed to pass the fields but assume they'll be figured out for you. However, you could perhaps allow both and then could leave this variable name alone. See my comment below.

dcicutils/ff_utils.py
```diff
@@ -457,65 +457,150 @@ def faceted_search(key=None, ff_env=None, item_type=None, **kwargs):
...
+                        if not qc_metric_list.get('qc_list'):
+                            continue
+                        for qc in qc_metric_list['qc_list']:
+                            qc_uuid = qc['value']['uuid']
+                            qc_info = {qc_uuid: {}}
+                            qc_meta = get_metadata(qc_uuid, key=key, ff_env=ff_env)
+                            qc_info[qc_uuid]['values'] = {k: v for k, v in qc_meta.items() if k not in ignored_qc_fields}
+                            qc_info[qc_uuid]['source_file_association'] = associated_file if associated_file != 'files' else 'raw_file'
+                            qc_info[qc_uuid]['source_file'] = entry['accession']
+                            qc_info[qc_uuid]['source_file_type'] = entry['file_type_detailed']
```
netsettler: It's a style issue, so not a big deal, but did you experiment with writing this in a way that is more visually clear, like this (untested):
```
qc_meta = get_metadata(qc_uuid, key=key, ff_env=ff_env)
qc_info = {
    qc_uuid: {
        'values': {k: v for k, v in qc_meta.items() if k not in ignored_qc_fields},
        'source_file_association': associated_file if associated_file != 'files' else 'raw_file',
        'source_file': entry['accession'],
        'soure_file_type': entry['file_type_detailed']
    }
}
```
The disadvantage is that if you have an error, it's a little harder to debug. Your way has a lot of friendly variables that can be seen separately when stepping. But the result is not as clear. A possible way to combine the two styles is to first compute these as local variables and then put it together:
```
values = ...
source_file_association = ...
...
qc_info = {
    qc_uuid: {
        'values': values,
       ...etc.
    }
}
```
I just think sometimes one sees big long columns of assignments like this and gets no easy sense of what's being done with it all. But I admit that's subjective.

dcicutils/ff_utils.py
```diff
@@ -457,65 +457,150 @@ def faceted_search(key=None, ff_env=None, item_type=None, **kwargs):
...
+                                'source_file_association': the file class (processed_file or raw_files)
+                                'source_file':, the accession of the file that the qc is linked to
+                                'source_file_type', the description of the file that the qc is linked to
+                                'experiment_description', the description of the experiment or experimentset
+                                'organism', the organism
+                                'experiment_type', the experiment type (in situ Hi-C, ChIP-seq)
+                                'experiment_subclass' (Hi-C)
+                                'source_experiment': the experiment the qc is linked to (if apply)
+                                'source_experimentSet': the experimentSet the qc is linked to
+                                'biosource_summary': the experiment biosource
```
netsettler: Something seems garbled here. Braces are unbalanced and in the wrong place. There's an extra empty string. Some commas in wrong places. I think you mean to say 
```
            {
                <qc_metric_uuid>: {
                    'values': <the values of the qc_metric object>,
                    'source_file_association': <the file class (processed_file or raw_files)>,
                    'source_file': <the accession of the file that the qc is linked to>,
                   ...etc.
                }
            }
```
I didn't fill them all out. You'll want to replace the "...etc." part. But I think the "<...>" notation is common for meta-syntactic identifiers that describe what goes in a position.

dcicutils/ff_utils.py
```diff
@@ -457,65 +457,167 @@ def faceted_search(key=None, ff_env=None, item_type=None, **kwargs):
...
     """
-    Given a uuid of an experiment set, return a dictionary of uuid : item
-    mappings of quality metric items.
+    Utility function to grap all the qc metrics from associated types of file such as:
+    'proccessed_files', 'other_processed_files', 'files'
+    inputs:
+        data: the metadata of a ExperimentSet or Experiment
+        associated_files: a list of the types of the files fields the qc metrics will be extracted from:
+            examples are = ['files', 'processed_files', 'other_processed_files']
+    Args:
```
willronchetti: I'd replace 'inputs' with 'Args' and delete the second heading. I get in principle what you're trying to do but I don't think its necessary

dcicutils/ff_utils.py
```diff
@@ -457,65 +457,167 @@ def faceted_search(key=None, ff_env=None, item_type=None, **kwargs):
...
+            exp_qc_metrics = fetch_files_qc_metrics(exp, associated_files, key=key, ff_env=ff_env)
+            if exp_qc_metrics:
+                for exp_qc_metric in exp_qc_metrics.values():
+                    exp_qc_metric['experiment_description'] = exp_description
+                    exp_qc_metric['organism'] = organism
+                    exp_qc_metric['experiment_type'] = experiment_type
+                    exp_qc_metric['experiment_subclass'] = experiment_subclass
+                    exp_qc_metric['source_experiment'] = exp['accession']
+                    exp_qc_metric['source_experimentSet'] = resp['accession']
+                    exp_qc_metric['biosource_summary'] = biosource_summary
```
willronchetti: I would consider following Kent's advice and writing this like so:
```
exp_qc_metric = {
    'experiment_description': exp_description
    'organism': organism
     ...
}

File: dcicutils/ff_utils.py [hunk 7]
```diff
@@ -457,65 +457,167 @@ def faceted_search(key=None, ff_env=None, item_type=None, **kwargs):
...
+                result.update(exp_qc_metrics)
+
+    description = resp.get('dataset_label', None)
+    ES_qc_metrics = fetch_files_qc_metrics(resp, associated_files, key=key, ff_env=ff_env)
+    if ES_qc_metrics:
+        for qc_metric in ES_qc_metrics.values():
+            qc_metric['experiment_description'] = description
```
willronchetti: Same comment as above here.

test/test_ff_utils.py
```diff
@@ -815,28 +828,64 @@ def test_faceted_search_users(integrated_ff):
     assert len(resp) == 10
 
 
+def test_fetch_qc_metrics_logic(mocked_replicate_experiment):
+    """
+    Tests only 'fetch_qc_metrics'
```
willronchetti: Maybe more specific docstring?

test/test_ff_utils.py
```diff
@@ -815,28 +828,64 @@ def test_faceted_search_users(integrated_ff):
...
+            "uuid": "7a2d8f3d-2108-4b81-a09e-fdcf622a0392",
+            "display_title": "QualityMetricPairsqc from 2018-04-27"
+        }
+        result = ff_utils.fetch_files_qc_metrics(mocked_replicate_experiment, ['processed_files'])
+        assert "7a2d8f3d-2108-4b81-a09e-fdcf622a0392" in result
+
+
+def test_get_qc_metrics_logic(mocked_replicate_experiment, qc_metrics):
+    """
+    End to end test on 'get_associated_qc_metrics'
```
willronchetti: Same as above - more info in docstring

PR #62, opened 2020-03-24
C4-90 Update utils for new production envs

willronchetti:
- Updates hard coded things to reflect the new (and old) reality
- This should be reviewed fairly thoroughly to be sure nothing is missed

### DIFF
--- a/dcicutils/beanstalk_utils.py
+++ b/dcicutils/beanstalk_utils.py
@@ -48,7 +48,7 @@ def delete_db(db_identifier, take_snapshot=True, allow_delete_prod=False):
         dict: boto3 response from delete_db_instance
     """
     # safety. Do not allow accidental programmatic deletion of webprod DB
-    if ('webprod' in db_identifier or 'production' in db_identifier) and not allow_delete_prod:
+    if 'prod' in db_identifier and not allow_delete_prod:
         raise Exception('Must set allow_delete_prod to True to delete RDS instance' % db_identifier)
     client = boto3.client('rds')
     timestamp = datetime.strftime(datetime.utcnow(), "%Y-%m-%d")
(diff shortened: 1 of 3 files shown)


dcicutils/beanstalk_utils.py
```diff
@@ -48,7 +48,7 @@ def delete_db(db_identifier, take_snapshot=True, allow_delete_prod=False):
         dict: boto3 response from delete_db_instance
     """
     # safety. Do not allow accidental programmatic deletion of webprod DB
-    if 'webprod' in db_identifier and not allow_delete_prod:
+    if ('webprod' in db_identifier or 'production' in db_identifier) and not allow_delete_prod:
```
netsettler: would it be safer to just do `'prod' in identifier`?

dcicutils/beanstalk_utils.py
```diff
@@ -246,7 +246,7 @@ def get_beanstalk_real_url(env):
     if env in urls:
         return urls[env]
 
-    if 'webprod' in env:
+    if 'webprod' in env or 'blue' in env or 'green' in env:
```
netsettler: This code is VERY fragile. Later we should change the environments to be `fourfront-prod-blue` and `fourfront-prod-green` so that blue/green can be used for other things (e.g., `fourfront-mastertest-blue` and `fourfront-mastertest-green`. This is OK for now, but I made a ticket to address it. Please add a comment saying something like:
```
#TODO (C4-91): reconsider environment names
```

dcicutils/ff_utils.py
```diff
@@ -22,6 +22,8 @@
 HIGLASS_BUCKETS = ['elasticbeanstalk-fourfront-webprod-wfoutput',
                    'elasticbeanstalk-fourfront-webdev-wfoutput']
 
+PRODUCTION_ENVS = ['fourfront-blue', 'fourfront-green']
```
netsettler: We did this in `fourfront` in `commands/create_mapping_on_deploy.py` and really need to centralize these. I started a PR for that and will try to follow up later. It's bad to have these in multiple places since if we update one and not another we'll get a mess. I made a ticket C4-92. Please add a TODO comment citing that ticket.

PR #63, opened 2020-03-25
Better handling of new environment names in s3utils

netsettler:
This addresses the problem in C4-95.

This also adds a `dcicutils.env_utils` with
* `is_cgap_env`
* `is_foufront_env`
* `is_stg_or_prd_env`

Bumps minor version to `0.10.0` for new functionality in addition to the bug fix.

### DIFF
--- a/dcicutils/env_utils.py
+++ b/dcicutils/env_utils.py
@@ -5,10 +5,18 @@ CGAP_STG_OR_PRD_NAMES = ['fourfront-cgap', 'fourfront-cgap-green', 'fourfront-cg
 
 
 def is_cgap_env(envname):
+    """
+    Returns True of the given string looks like a CGAP elasticbeanstalk environment name.
+    Otherwise returns False.
+    """
     return 'cgap' in envname
 
 
 def is_fourfront_env(envname):
+    """
+    Returns True of the given string looks like a Fourfront elasticbeanstalk environment name.
+    Otherwise returns False.
+    """
     return 'cgap' not in envname
 
 
--- a/dcicutils/env_utils.py
+++ b/dcicutils/env_utils.py
@@ -13,6 +13,12 @@ def is_fourfront_env(envname):
 
 
 def is_stg_or_prd_env(envname):
+    """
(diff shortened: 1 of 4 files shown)


dcicutils/env_utils.py
```diff
@@ -0,0 +1,21 @@
+FOURFRONT_STG_OR_PRD_TOKENS = ['webprod', 'blue', 'green']
+FOURFRONT_STG_OR_PRD_NAMES = ['staging', 'stagging', 'data']
+CGAP_STG_OR_PRD_TOKENS = []
+CGAP_STG_OR_PRD_NAMES = ['fourfront-cgap', 'fourfront-cgap-green', 'fourfront-cgap-blue']
```
willronchetti: Hopefully these can be `cgap-green` and `cgap-blue` but I suspect we will run into the hard coding problem again if that change is made (to be done later).

dcicutils/env_utils.py
```diff
@@ -0,0 +1,22 @@
...
+
+def is_cgap_env(envname):
+    return 'cgap' in envname
+
+
+def is_fourfront_env(envname):
+    return 'cgap' not in envname
+
+
+def is_stg_or_prd_env(envname):
```
willronchetti: Self documenting so probably not worth adding a docstring (definitely not for the above 2 functions).

dcicutils/env_utils.py
netsettler: Yeah, I figured when we come up with an official new naming scheme we can fix it here. This gives us a little flex to experiment in the old paradigm.

dcicutils/env_utils.py
netsettler: Oh, yeah, good point. But indeed, I worked hard on the name because I started the name as just "_prd_" or "_production_" and was worried there were too many interpretations. Whatever the doc string, the name itself has to be crystal clear. I'll still put in a doc string.

netsettler: OK. I added doc strings.

netsettler:
Adjustment to hopefully fix a problem in foursight.

PR #66, opened 2020-03-28
More backported deployment stuff for env_utils and beanstalk_utils

netsettler:
I've implemented a number of operations that were previously managed in fourfront, so that I can have common implementations used by both cgap and fourfront.

**beanstalk_utils**

* source_beanstalk_env_vars

**env_utils**

* All of the variables `ENV_HOTSEAT`, etc. which are here `FF_ENV_HOTSEAT` or `CGAP_ENV_HOTSEAT` but which largely should not be needed in Fourfront because I added other abstractions.
* `BEANSTALK_PROD_MIRRORS` and `BEANSTALK_TEST_ENVS` contain both CGAP and FF values, but also shouldn't be needed outside of this module.
*  New functions `is_test_env`, `is_hotseat_env` that are predicates on an environment name. These operate at a better level of abstraction for Fourfront use.
* Function `get_mirror_env_from_context` (able to get an environment name and its mirror name from a config file and optionally overridden by an environment variable) was moved from Fourfront, where it was called `get_mirror_env`. Function `get_env_from_context` is new and works similarly.
* New function `guess_mirror_env` that does heuristic (but more regular) guessing about what the mirror env should look like.

**Design Issues**

 * Some question about whether `guess_mirror_env` or `get_mirror_env_from_context` is the right thing to use outside this module. But this creates some vocabulary to talk about that.

* There could be some reasonable debate over what the right default setting is four allow_environ in the `get_..._env` functions. We mentioned this on Slack earlier this evening so I implemented the option just in case. I kind of like it, but I don't want any surprises so I didn't turn it on by default.

### DIFF
--- a/dcicutils/env_utils.py
+++ b/dcicutils/env_utils.py
@@ -1,11 +1,7 @@
 import os
 
 
-FOURFRONT_STG_OR_PRD_TOKENS = ['webprod', 'blue', 'green']
-FOURFRONT_STG_OR_PRD_NAMES = ['staging', 'stagging', 'data']
-CGAP_STG_OR_PRD_TOKENS = []
-CGAP_STG_OR_PRD_NAMES = ['fourfront-cgap', 'fourfront-cgap-green', 'fourfront-cgap-blue']
-
+FF_ENV_DEV = 'fourfront-dev'  # Maybe not used
 FF_ENV_HOTSEAT = 'fourfront-hotseat'
 FF_ENV_MASTERTEST = 'fourfront-mastertest'
 FF_ENV_PRODUCTION_BLUE = 'fourfront-blue'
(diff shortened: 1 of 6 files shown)


dcicutils/env_utils.py
```diff
@@ -47,3 +99,48 @@ def is_stg_or_prd_env(envname):
...
+    if allow_environ:
+        environ_env_name = os.environ.get('ENV_NAME')
+        if environ_env_name:
+            return environ_env_name
+    return settings.get('env.name')
+
+
+def get_mirror_env_from_context(settings, allow_environ=False, allow_guess=True, ):
+    # TODO: I added allow_environ featurism here but did not yet enable it.
+    #       Want to talk to Will about whether we should consider that a compatible or breaking hcange.
```
willronchetti: This sounds fine to me, it would be nice though if we could enable allow_environ with an environment variable (I know that sounds weird) because we could then pull settings from Elastic Beanstalk templates.

netsettler: `allow_environ` only turns on whether it looks for the environment variable. If the environment variable is not there, then it acts as if you want it to use declaration files or other means. The order is:
* Look at environment variable for a mirror environment (if allow_environ is True)
* Look in config.
* Look for an environment variable for regular env and guess mirror based on that (if allow_environ is True and allow_guess is True)
* Look for regular env in config and guess mirror based on that (if allow_environ is True and allow_guess is True)
If you want all four tests to be done that way, I would be OK with that.

netsettler:
And I plan to release this as 0.12.0, not beta, so I can make snovault do a `>=' test.

dcicutils/env_utils.py
```diff
@@ -1,8 +1,60 @@
...
+FF_ENV_WEBPROD = 'fourfront-webprod'
+FF_ENV_WEBPROD2 = 'fourfront-webprod2'
+FF_ENV_WOLF = 'fourfront-wolf'
+
+CGAP_ENV_HOTSEAT = 'fourfront-cgaphot'
+CGAP_ENV_MASTERTEST = 'fourfront-cgaptest'
+CGAP_ENV_PRODUCTION_BLUE = 'fourfront-cgapblue'
+CGAP_ENV_PRODUCTION_GREEN = 'fourfront-cgapgreen'
+CGAP_ENV_STAGING = 'fourfront-cgapstaging'
+CGAP_ENV_WEBDEV = 'fourfront-cgapdev'
```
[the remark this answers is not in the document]
netsettler: I can rename it, but the name isn't used for much. `CGAP_ENV_WEBDEV` is in `BEANSTALK_TEST_ENVS` below. Are you suggesting I here call it `CGAP_ENV_DEV` _and_ remove it from `BEANSTALK_TEST_ENVS`?

dcicutils/env_utils.py
```diff
@@ -1,8 +1,60 @@
...
 
+FF_ENV_HOTSEAT = 'fourfront-hotseat'
+FF_ENV_MASTERTEST = 'fourfront-mastertest'
+FF_ENV_PRODUCTION_BLUE = 'fourfront-blue'
+FF_ENV_PRODUCTION_GREEN = 'fourfront-green'
+FF_ENV_STAGING = 'fourfront-staging'
+FF_ENV_WEBDEV = 'fourfront-webdev'
+FF_ENV_WEBPROD = 'fourfront-webprod'
+FF_ENV_WEBPROD2 = 'fourfront-webprod2'
+FF_ENV_WOLF = 'fourfront-wolf'
```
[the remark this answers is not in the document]
netsettler: Yes, this was just in case. I thought about marking it `# unused` here, but then if we started to use it we'd have to make a code change. :(

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "dcicutils"
-version = "0.11.0"
+version = "0.12.0b1"
```
netsettler: I plan to change this to `0.12.0` before merging.

dcicutils/env_utils.py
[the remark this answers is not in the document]
willronchetti: I would rename to CGAP_ENV_DEV only.

dcicutils/env_utils.py
willronchetti: Makes sense.

willronchetti:
Commenting approval of latest changes, looks good.

PR #68, opened 2020-03-31
CGAP bucketing fixes (C4-92, C4-102)

netsettler:
I'm hoping this will fix some or all of Soo's reported problems with access. I wrote some tests that failed in a way similar to what I think was described before making the changes, and this fixes those cases. There could be other issues. 

There are a couple of places I didn't fix here that I have flagged in comments so that they'll be reviewable. We may need to add those, too, or maybe we can add them in a later pass.


dcicutils/beanstalk_utils.py
```diff
@@ -206,24 +209,36 @@ def swap_cname(src, dest):
...
+        if env.get('CNAME') == magic_cname:
             # we found data
             return env.get('EnvironmentName')
 
 
+def compute_ff_prd_env():
+    return _compute_prd_env_for_project('ff')
+
+
+whodaman = compute_ff_prd_env  # This naming is deprecated but retained for compatibility.
```
willronchetti: We can deprecate this function elsewhere as part of a compatibility patch

dcicutils/ff_utils.py
```diff
@@ -20,11 +20,13 @@
     from urllib.parse import urlencode
 
 
+# TODO (C4-92, C4-102): Probably to centralize this information in env_utils. Also figure out relation to CGAP.
 HIGLASS_BUCKETS = ['elasticbeanstalk-fourfront-webprod-wfoutput',
                    'elasticbeanstalk-fourfront-webdev-wfoutput']
 
 
 # TODO (C4-92): Centralize this information, it is repeated in other repos
+# TODO (C4-102): Does this need to include CGAP envs? As part of the same list, or as a separate list?
```
willronchetti: It may make sense to add cgap as this package can be used on CGAP as well as Fourfront despite its name, though that may lead to some interesting complications.

dcicutils/qa_utils.py
```diff
@@ -0,0 +1,20 @@
+"""
+qa_utils: Tools for use in quality assurance testing.
+"""
+
+from .misc_utils import PRINT
+
+
+def mock_not_called(name):
```
willronchetti: Why not `assert not mock.called, "mock called where not expected"` ? Is the point that you preserve args/kwargs?

test/test_beanstalk_utils.py
```diff
@@ -1,34 +1,134 @@
...
-    with mock.patch('dcicutils.beanstalk_utils.beanstalk_info') as man_not_hot:
-        man_not_hot.return_value = {'CNAME': 'take-of-your-jacket'}
-        url = bs.get_beanstalk_real_url('take-of-your-jacket')
-        assert url == 'http://take-of-your-jacket'
 
+def test_get_beanstalk_real_url_fake_staging():
+    with mock.patch('dcicutils.beanstalk_utils._compute_prd_env_for_project') as mock_whodaman:
+        with mock.patch('dcicutils.beanstalk_utils.beanstalk_info') as mock_beanstalk_info:
+            mock_whodaman.return_value = 'webprod-2'
+            mock_beanstalk_info.side_effect = _mocked_beanstalk_info  # mock_not_called('dcicutils.beanstalk_utils.beanstalk_info')
```
willronchetti: Is this comment meant to be on every test here?

dcicutils/qa_utils.py
netsettler: Short answer is: personal style. But the style isn't entirely arbitrary in choice. I find those mock properties to be easy to make typos on. I've misspelled them and found it had ill effect. e.g., suppose you do "if mock.callled" or something like that, what ends up happening is it returns a magic mock object that counts as true. So I tend to not rush to use those things unless the thing is hard to mock otherwise. Plus you don't have to change the structure of the code if you decide adjust what the function does, whereas if you want to go from "not mock.called" to doing something slightly different, you have to write not just the changed functionality but change the test.

PR #70, opened 2020-04-12
Misc small fixes to env_utils, beanstalk_utils, s3_utils (C4-34)

netsettler:
This started out to be some changes for `env_utils` in support of CGAP in Poetry (C4-34), but it turned out I did not need the changes I was going to do for that. Still, there were some tests that were broken by the recent Fourfront cname swap (that were really because the tests were written badly, not because any functionality was broken). So if there is any "real" motivation for this PR, it would be that.

There are also a number of opportunistic changes riding along on this PR just because I made them as I saw them. Full list of changes:

* Make various `env_utils` predicates (`is_cgap_env`, etc.) allow `None` harmlessly, returning `False`. (Better than getting a ValueError for using `in` on wrong type data.)
* Rename `guess_mirror_env` to `get_standard_mirror_env`, since it's not really guessing. (I kept the old name for compatibility, but the doc string notes it's deprecated. Maybe there's a way to actively tell the user, but if so I didn't do that.)
* Add `make lint` to `Makefile`
* Add function `ignored` to `misc_utils`. This is to declare unused arguments as something lint tools shouldn't fuss over.
* Fix tests in `beanstalk_utils` that were broken by the cname swap; that is, they never were the right test. But now they are.
* Add some additional tests in `env_utils`, `beanstalk_utils`, `s3_utils` (in addition to appropriate testing for newly created functionality).
* Minor version bump for new functionality.

### DIFF
--- a/dcicutils/env_utils.py
+++ b/dcicutils/env_utils.py
@@ -164,7 +164,7 @@ def is_cgap_env(envname):
     Returns True of the given string looks like a CGAP elasticbeanstalk environment name.
     Otherwise returns False.
     """
-    return 'cgap' in envname
+    return 'cgap' in envname if envname else False
 
 
 def is_fourfront_env(envname):
@@ -172,7 +172,7 @@ def is_fourfront_env(envname):
     Returns True of the given string looks like a Fourfront elasticbeanstalk environment name.
     Otherwise returns False.
     """
-    return 'cgap' not in envname
+    return 'cgap' not in envname if envname else False
 
 
 def is_stg_or_prd_env(envname):
@@ -182,6 +182,8 @@ def is_stg_or_prd_env(envname):
     True whether or not they are actually the currently live instance. (This function doesn't change its
     state as blue or green is deployed, in other words.)
     """
+    if not envname:
+        return False
     stg_or_prd_tokens = CGAP_STG_OR_PRD_TOKENS if is_cgap_env(envname) else FOURFRONT_STG_OR_PRD_TOKENS
     stg_or_prd_names = CGAP_STG_OR_PRD_NAMES if is_cgap_env(envname) else FOURFRONT_STG_OR_PRD_NAMES
     if envname in stg_or_prd_names:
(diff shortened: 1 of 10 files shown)


dcicutils/env_utils.py
```diff
@@ -234,6 +233,34 @@ def get_mirror_env_from_context(settings, allow_environ=ALLOW_ENVIRON_BY_DEFAULT
...
+    This function knows about the standard mirroring rules and infers a mirror env only from that.
+    (In tha sense, it is not guessing and probably needs to be renamed.)
+    If there is no mirror, it returns None.
+
+    This is not the same as blue_green_mirror_env(envname), which is purely syntactic.
+
+    This is also not the same as get_mirror_env_from_context(...), which infers the mirror from contextual
+    information such as config files and environment variables.
+    """
+    return BEANSTALK_PROD_MIRRORS.get(envname, None)
```
willronchetti, line 247: Should `envname` be `blue` or `green` in this case? ie: not `data` or `staging` but `mastertest` is okay?

test/test_beanstalk_utils.py
```diff
@@ -114,38 +122,140 @@ def communicate(self):
                     'AWS_FAKE_SECRET': 'amazon'
                 }
 
-def test_magic_cnames():
+
+def _ip_addresses(hostname):
+    return sorted(socket.gethostbyname_ex(hostname)[2])
+
+
+def test_magic_cnames_by_production_ip_address():
```
willronchetti, line 130: I get that this should always work, but does it really belong in a unit test here?

netsettler: If's a form of consistency checking. Whether we should reach out to the net to test it is another question but otherwise it's just a magic constant.

If we had a rule against there being unit tests that talked to the net, I'd move it to integration tests but it seems to me OK here for now.

dcicutils/env_utils.py
netsettler: I think you're asking for better doc on what kinds of arguments this wants. I can add that.

netsettler: Done.

PR #71, opened 2020-04-15
Add dcicutils.misc_utils.get_setting_from_context

netsettler:
This adds support for a customization pattern we expect we'll reuse. See the doc string for `get_setting_from_context` for information on what it does.

### DIFF
--- a/dcicutils/env_utils.py
+++ b/dcicutils/env_utils.py
@@ -1,5 +1,5 @@
 import os
-
+from .misc_utils import get_setting_from_context
 
 FF_ENV_DEV = 'fourfront-dev'  # Maybe not used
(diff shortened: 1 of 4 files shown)


dcicutils/env_utils.py
```diff
@@ -210,14 +210,10 @@ def is_hotseat_env(envname):
 
 
 def get_env_from_context(settings, allow_environ=ALLOW_ENVIRON_BY_DEFAULT):
```
willronchetti, line 212: I would add a comment here for the next major version to remove this argument, as we don't want to make a potentially breaking change now but probably should in the future.

netsettler:
Oops. I thought I had merged this. I apparently tagged the branch. This merge will make things right.

PR #74, opened 2020-04-27
C4-137 Provision Indexer Deployment

willronchetti:
- This PR is a draft - this approach should be reviewed with Kent.
- Adds some functionality to `build_ini_stream_from_template` to pick up the `ENCODED.INDEXER` option either from environment variable or argument
- When it encounters this, it adds an option to the `ini` file that will tell the application it is running just the indexer (it should be visible in the registry settings).
- When you deploy an indexer, it's `bs_name` is the name normally passed when deploying the corresponding environment. 
- So to deploy an indexer, you simply set an environment variable on the Beanstalk Console called `ENCODED.INDEXER`.
- Also includes `VirtualApp`, to replace `TestApp` in our applications


test/test_deployment_utils.py
```diff
@@ -277,6 +277,43 @@ def now(cls):
...
+
+                # for this test, we're going to pretend we deployed with bs_name == 'fourfront-indexer',
+                # which should throw an exception
+                def mocked_os_get(val, default):
+                    if val == 'ENCODED_BS_ENV':
+                        return 'fourfront-indexer'
+                    else:
+                        return default
+
+                with mock.patch("os.environ.get", side_effect=mocked_os_get):
```
netsettler, line 312: This won't consistently change `x in os.environ` or `os.environ[x]`. I think it's better to bind the individual values by doing
```
from qa_utils import override_environ
...
with override_environ(ENCODED_BS_ENV='fourfront-indexer'):
    with pytest.raises(RuntimeError):
        TestDeployer.build_ini_file_from_template(...)
```

test/test_deployment_utils.py
```diff
@@ -277,6 +277,43 @@ def now(cls):
...
+            mock_exists.return_value = True
+            with mock.patch("io.open", side_effect=mocked_open):
+
+                # for this test, we're going to pretend we deployed with bs_name == 'fourfront-indexer',
+                # which should throw an exception
+                def mocked_os_get(val, default):
+                    if val == 'ENCODED_BS_ENV':
+                        return 'fourfront-indexer'
+                    else:
+                        return default
```
netsettler, line 310: This is going to return default for every other value, simulating that no other variable is present. Is that what you want? Anyway, the technique I show below is more general and lets you more easily control just those variables you want, leaving others alone.

dcicutils/deployment_utils.py
```diff
@@ -39,12 +39,13 @@ class Deployer:
 
     TEMPLATE_DIR = None
     INI_FILE_NAME = "production.ini"
+    INDEXER_ENTRY = 'ENCODED.INDEXER = "true"\n'
```
netsettler, line 42: I think the entries are all lowercase. I think the `.ini` file entry should be `encoded.indexer` or even just `indexer` given that the namespace is already encoded.

willronchetti:
Kents pulling this down and making some changes, in the meantime I've put up a first draft of `VirtualApp`. There are a lot of things I think we should do with it, which I will probably implement tomorrow.

dcicutils/misc_utils.py
```diff
@@ -11,6 +13,54 @@
...
+    logging.basicConfig()
+
+
+    def __init__(self, app, environ):
+        """ Builds an encoded application, allowing you to submit requests to an encoded application
+
+        :param app: return value of get_app(config_uri, app_name)
+        :param environ: options to pass to the application. Usually permissions.
+        """
+        self.virtual_app = TestApp(app, environ)
```
netsettler, line 29: If you're going to make it have an attribute that holds a TestApp (and that's not a terrible plan, just a design choice), then you don't want to also inherit from it. In that case, you'd be the instance itself, inheriting its `__init__` method from `TestApp` so it doesn't need a new one unless it's going to rewrite the instantiation protocol.

dcicutils/misc_utils.py
```diff
@@ -11,6 +13,54 @@
...
+
+    def get(self, url, **kwargs):
```
netsettler, line 31: If you inherit from `TestApp`, the way you write this is:
```
def get(self, url, **kwargs):
    """Wrapper for TestApp...etc."""
    logging.info("OUTGOING HTTP GET: %s' % url)
    super(VirtualApp, self).get(url, **kwargs)
```
The syntax may have a simpler option in Python3, but my point is you just call the super rather than calling something in a slot.

dcicutils/misc_utils.py
```diff
@@ -11,6 +13,54 @@
 PRINT = print
 
 
+class VirtualApp(TestApp):
```
netsettler, line 16: If you're going to inherit from `TestApp`, you don't need to put a TestApp also in the `.virtual_app` slot below. That's an either/or design choice.

PR #79, opened 2020-05-02
Support for 'make publish' (C4-150)

netsettler:
Following up on the auto-publishing problem described in [C4-150](https://hms-dbmi.atlassian.net/browse/C4-150), this adds a manual workaround to the publication problem.

* `scripts/publish` is the same script as `[python_lambda]bin/publish`, which was written in a way that can be run from any directory containing poetry information in `pyproject.toml`.
* `Makefile` adds a `make` target that will call the script.
* Bumping the patch version only because this is visible only to maintainers of this library, not to users, and is really a kind of bug fix.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,9 @@ test:
 update:  # updates dependencies
 	poetry update
 
+publish:
+	scripts/publish
+
 help:
 	@make info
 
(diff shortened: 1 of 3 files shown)


netsettler:
@willronchetti, just to keep things moving along before our meeting tomorrow, I'm going to merge this. Please review after-the-fact. I think it's pretty non-controversial and it's already tested in another repo.

PR #81, opened 2020-05-04
Misc clarifications in deployment_utils

netsettler:
The main purposes of this PR are these:

* Fix a bug in the defaulting of the `indexer` value in the `main` program, which was defaulting to `False` (which in turn hides the environment value of `True` that was trying to get through). I changed the default to None from False.  The logic to turn an unspecified value to False happens elsewhere already.

* Add an accessor `dcicutils.deployment_utils.boolean_setting` which can be used to recover a boolean from settings that have `"true"` and `"false"` values floating around.

* Added unit tests.

Opportunistic:

* I added some function documention some of the new `lang_utils` functions and changed `relative_time_string` to allow it to take a `datetime.timedelta` as an argument. (It's easier to see such needs when writing doc.)

### DIFF
--- a/dcicutils/deployment_utils.py
+++ b/dcicutils/deployment_utils.py
@@ -18,20 +18,45 @@ For example, fourfront/deploy/generate_production_ini.py might contain:
 
 """
 
+import argparse
 import datetime
 import glob
 import io
 import json
+import logging
 import os
 import pkg_resources
 import re
 import subprocess
 import sys
 import toml
-import argparse
 
-from dcicutils.env_utils import get_standard_mirror_env, data_set_for_env, get_bucket_env, INDEXER_ENVS
-from dcicutils.misc_utils import PRINT
+from .env_utils import get_standard_mirror_env, data_set_for_env, get_bucket_env, INDEXER_ENVS
+from .misc_utils import PRINT
+
+
+def boolean_setting(settings, key, default=None):
+    """
+    Given a setting from a .ini file and returns a boolean interpretation of it.
+    - Treats an actual python boolean as itself
+    - Using case-insensitive comparison, treats the string 'true' as True and both 'false' and '' as False.
+    - If an element is missing,
+    - Treats None the same as the option being missing, so returns the given default.
+    - Raises an error for any non-standard value. This is harsh, but it should not happen.
+    """
+    if key not in settings:
+        return default
+    setting = settings[key]
+    if isinstance(setting, str):
+        setting_lower = setting.lower()
+        if setting_lower in ("", "false"):
+            return False
+        elif setting_lower == "true":
+            return True
+        else:
+            return setting
+    else:  # booleans, None, and odd types (though we might want to consider making other types an error).
+        return setting
 
 
 class Deployer:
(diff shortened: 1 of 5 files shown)


dcicutils/deployment_utils.py
```diff
@@ -287,11 +312,12 @@ def main(cls):
                                 help="an ElasticSearch namespace",
                                 default=None)
             parser.add_argument("--indexer",
-                                help="whether this system does indexing",
-                                action='store_true',
-                                default=False)
+                                help="whether this server does indexing at all",
+                                choices=["true", "false"],
+                                default=None)
```
willronchetti, line 317: Is None really what we want here? Will it pick up a default option elsewhere? Same question for `--index_server`, I think default should be `False` in that case.

test/test_deployment_utils.py
```diff
@@ -684,3 +684,70 @@ def check(self, line):
                                es_server="search-fourfront-webdev-5uqlmdvvshqew46o46kcc2lxmy.%s" % us_east,
                                line_checker=Checker(expect_indexer=None,
                                                     expect_index_server="true"))
+
+
+def test_deployment_utils_main():
+
+    fake_template = "something.ini"
+    with override_environ(ENV_NAME='fourfront-foo'):
```
willronchetti, line 692: Test could use some documentation I'd say.

dcicutils/deployment_utils.py
```diff
@@ -18,20 +18,45 @@ def main():
...
+    setting = settings[key]
+    if isinstance(setting, str):
+        setting_lower = setting.lower()
+        if setting_lower in ("", "false"):
+            return False
+        elif setting_lower == "true":
+            return True
+        else:
+            return setting
+    else:  # booleans, None, and odd types (though we might want to consider making other types an error).
```
willronchetti, line 58: Should we not coerce None into False?

dcicutils/deployment_utils.py
```diff
@@ -18,20 +18,45 @@ def main():
...
+        return default
+    setting = settings[key]
+    if isinstance(setting, str):
+        setting_lower = setting.lower()
+        if setting_lower in ("", "false"):
+            return False
+        elif setting_lower == "true":
+            return True
+        else:
+            return setting
```
willronchetti, line 57: Should we not coerce whatever this could be to False?

dcicutils/deployment_utils.py
netsettler: Yes, the problem is that we either have to have a rule that the command arguments override the environment variables or vice versa. The rule was (and I guess I could change this because it isn't clear it's right) that the arguments override the env variables, and so putting False here made the env variables not work. But anyway there is logic that picks up the env variables and defaults it. (This is all heavily unit tested but could still have gone by you as even then it's subtle.)

Another complexity in the design space is that env variables are strings and command arguments can be real datatypes. That's weird because when the env variables are merged, you also have to do various kinds of coercions to even get them on a level playing field, and THEN you have to flatten them again to strings to put them in an ini file and AGAIN recover them as strings later. I REALLY wish we were using a json structure to hold this data so we didn't have to do that second coerce-to-string.

test/test_deployment_utils.py
netsettler: OK. Will add.

dcicutils/deployment_utils.py
netsettler: Since None is already a false value, I thought it might be better to leave it as None in case we have to debug the difference between an explicit value and no value. I could change that if you thought it wise ,but my sense was it was better to leave the thin distinction.

dcicutils/deployment_utils.py
netsettler: Again, it already functions as a false value. I really wanted to raise an error here, but was nervous about where that error would be perceived.  Returning a stray value might make it easier to spot the error downstream than if we coerce it.

dcicutils/deployment_utils.py
willronchetti: Yeah I think that's a good point, so let's keep it.

dcicutils/deployment_utils.py
willronchetti: Again seems reasonable.

dcicutils/deployment_utils.py
netsettler: I'm leaving this alone for now.

PR #84, opened 2020-05-08
Helping avoid unforced errors

netsettler:
It often happens that flake8 fusses over weird things. That turns out to make one do an impulse decision to just believe it. But that's an accident waiting to happen because it complains about unused variables and pytest requires that fixtures look like unused variables. So I made a way to declare them. See the doc string for `dcicutils.qa_utils.notice_pytest_fixtures`, which has a detailed example in the doc string. The function itself is utterly trivial, but the thing it implements is actually quite important to avoiding errors, and the test case really does properly test that this will both placate flake8 and not interfere with fixtures.

### DIFF
--- a/test/test_qa_utils.py
+++ b/test/test_qa_utils.py
@@ -3,11 +3,19 @@ import os
 import pytest
 import pytz
 import re
+import subprocess
 import time
-import unittest
 import uuid
 
 from dcicutils.qa_utils import mock_not_called, local_attrs, override_environ, ControlledTime
+from unittest import mock
+
+# The following line needs to be separate from other imports. It is PART OF A TEST.
+from dcicutils.qa_utils import notice_pytest_fixtures   # Use care if editing this line. It is PART OF A TEST.
+from .fixtures.sample_fixtures import MockMathError, MockMath, math_enabled
+
+
+notice_pytest_fixtures(math_enabled)   # Use care if editing this line. It is PART OF A TEST.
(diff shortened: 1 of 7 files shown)


test/test_qa_utils.py
```diff
@@ -3,11 +3,19 @@
...
 
 from dcicutils.qa_utils import mock_not_called, local_attrs, override_environ, ControlledTime
+from unittest import mock
+
+# The following line needs to be separate from other imports. It is PART OF A TEST.
+from dcicutils.qa_utils import notice_pytest_fixtures   # Use care if editing this line. It is PART OF A TEST.
+from .fixtures.sample_fixtures import MockMathError, MockMath, math_enabled
+
+
+notice_pytest_fixtures(math_enabled)   # Use care if editing this line. It is PART OF A TEST.
```
willronchetti, line 18: So is this what you would do at the top of a test file if you want the "unused" errors to go away?

netsettler: Yep. I think it effectively says that in the doc string. Let me know if you think more is needed. Really you can do it for things that don't have unused things, too. It would be nice to know what fixtures are in use. But I think almost all fixtures will in fact look globally unused, so you'd be well-motivated. It's harder to tell within a function because some things have fixture effects but they also get used.

willronchetti: Makes sense to me.

PR #86, opened 2020-05-15
C4-58 Deployment APIs

willronchetti:
- Adds a Python 3.7 build to Travis so we can "support it".
- Allows Python 3.7 for versions after this one that way 3.7 users do not have outdated versions of the library.
- Adds deployment APIs to dcicutils for both general use and indexer-environment use.
- The intention is to use these new utilities to provide powerful deployment functionality in Foursight.
- This has gone through integrated testing but as of right now there are no unit tests. Advice on this matter would be helpful as I have written this in a way that I think is hard to do (non-integrated) unit testing. For now it is marked as WIP until this issue is resolved adequately.


.travis.yml
```diff
@@ -12,25 +8,26 @@ env:
...
-      provider: script
-      script: poetry publish --build --username=$PYPI_USER --password=$PYPI_PASSWORD
-      on:
-        tags: true
-    after_success:
-    - echo "Finished running tests"
-    - |
+    - python: "3.6"
+    - python: "3.7"
+before_install:
```
netsettler, line 13: This is a different indentation level. Is that right? I don't know the yaml rules for this, but it doesn't look right.

dcicutils/deployment_utils.py
```diff
@@ -58,7 +64,282 @@ def boolean_setting(settings, key, default=None):
...
+        # XXX: Make worker tier perhaps?
+        for idx, option in enumerate(configuration):
+
+            # make it a big machine (or not if we say so)
+            if option['OptionName'] == 'InstanceType':
+                option['Value'] = cls.DEFAULT_INDEXER_SIZE if not size else size
+
+            # add the additional env variable here as well
+            if option['OptionName'] == 'EnvironmentVariables':
+                option['Value'] += ',ENCODED_INDEX_SERVER=True'
```
netsettler, line 237: Or (untested):
```
adjustments = {
    'InstanceType': lambda value: cls.DEFAULT_INDEXER_SIZE if value is None else value,
    'EnvironmentVariables': lambda value: value + ",ENCODED_INDEX_SERVER=True'.
}
for option in configuration:
   option['Value'] = adjustments.get(option['OptionName'], lambda x: x)(option['Value'])
```

netsettler: Either way, is comma really the right separator for `EnvironmentVariables`?

dcicutils/deployment_utils.py
```diff
@@ -58,7 +64,282 @@ def boolean_setting(settings, key, default=None):
...
+class EBDeployer:
+
+    S3_BUCKET = 'dcic-application-versions'
+    EB_APPLICATION = '4dn-web'  # XXX: will need to change if this changes -Will
+    EB_CONFIGURATION_SETTINGS = 'ConfigurationSettings'
+    EB_OPTION_SETTINGS = 'OptionSettings'
+    EB_ENV_VARIABLE_NAMESPACE = 'aws:elasticbeanstalk:application:environment'
+    DEFAULT_INDEXER_SIZE = 'c5.4xlarge'
+    UPDATING = 'Updating'
+    STATUS = 'Status'
```
netsettler, line 76: If these were at toplevel instead of in the class, you'd not have to refer to them with `cls.` prefixes. In terms of modularity, that's a little less connected, but really you're not relying on these names for their class-oriented definition and also you don't want anyone setting the value of these in the class.

dcicutils/deployment_utils.py
```diff
@@ -58,7 +64,282 @@ def boolean_setting(settings, key, default=None):
...
+                TemplateName=CGAP_ENV_INDEXER,
+                VersionLabel=app_version
+            )[cls.STATUS] == cls.UPDATING and cls.delete_indexer_template(eb_client, FF_ENV_INDEXER)
+        elif is_fourfront_env(env_name):
+            return eb_client.create_environment(
+                ApplicationName=cls.EB_APPLICATION,
+                EnvironmentName=FF_ENV_INDEXER,
+                TemplateName=FF_ENV_INDEXER,
+                VersionLabel=app_version
+            )[cls.STATUS] == cls.UPDATING and cls.delete_indexer_template(eb_client, FF_ENV_INDEXER)
```
netsettler, line 285: As noted above, this would be prettier and less confusing if `cls.` were not interfering.

dcicutils/deployment_utils.py
```diff
@@ -58,7 +64,282 @@ def boolean_setting(settings, key, default=None):
...
+
+    @classmethod
+    def delete_indexer_template(cls, client, template_name):
+        """ Wrapper for "delete_configuration_template" that will only accept indexer_envs
+
+        :param client: boto3 elasticbeanstalk client
+        :param template_name: template to delete, will only accept indexer env templates
+        :return: True in success, False otherwise
+        """
+        if template_name not in [FF_ENV_INDEXER, CGAP_ENV_INDEXER]:
```
netsettler, line 297: It would be nice to have an `is_indexer_env` in `env_utils` rather than doing this at the naming level.

dcicutils/deployment_utils.py
netsettler: Really if there were an `indexer_name_for_env` to get the client name. These clauses seem otherwise the same and could be collapsed as they do the same thing.

dcicutils/deployment_utils.py
```diff
@@ -58,7 +64,282 @@ def boolean_setting(settings, key, default=None):
...
+        eb_client = boto3.client('elasticbeanstalk')
+        if is_cgap_env(env_name):
+            return eb_client.create_environment(
+                ApplicationName=cls.EB_APPLICATION,
+                EnvironmentName=CGAP_ENV_INDEXER,
+                TemplateName=CGAP_ENV_INDEXER,
+                VersionLabel=app_version
+            )[cls.STATUS] == cls.UPDATING and cls.delete_indexer_template(eb_client, FF_ENV_INDEXER)
+        elif is_fourfront_env(env_name):
+            return eb_client.create_environment(
```
netsettler, line 280: Does `create_environment` not make an environment or does this just take a while and not return before it can ask for the `cls.STATUS`?

dcicutils/deployment_utils.py
```diff
@@ -58,7 +64,282 @@ def boolean_setting(settings, key, default=None):
...
+    def deploy_new_version(cls, env_name, path_to_repo, application_version_name):
+        """  Deploys a new version to EB env_name, where env_name must be one of the valid environments
+
+        :param env_name: env to deploy to
+        :param path_to_repo: path to repo to deploy
+        :param application_version_name: application version name to give this version
+        :raises: RuntimeError if repo does not match env
+        """
+        if is_fourfront_env(env_name):
+            if 'fourfront' not in path_to_repo:
```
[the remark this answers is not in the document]
willronchetti: I'll take just the beginning `/` just because you could specify a path like `../fourfront`, which I think should work.

dcicutils/deployment_utils.py
```diff
@@ -58,7 +64,282 @@ def boolean_setting(settings, key, default=None):
...
+        elif is_cgap_env(env_name):
+            if 'cgap-portal' not in path_to_repo:
+                raise RuntimeError('Tried to deploy cgap env but path to repo does not contain "cgap-portal",'
+                                   ' aborting: %s' % path_to_repo)
+            cls._deploy(application_version_name, env_name)
+        else:
+            raise RuntimeError('Tried to deploy to invalid environment: %s' % env_name)
+
+    @staticmethod
+    def extract_environment_id(env_name):
```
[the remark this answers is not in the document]
willronchetti: Maybe, but it requires an API call to AWS which makes it 'feel' different.

dcicutils/deployment_utils.py
willronchetti: It does, but it doesn't block as it would take a long time. It should transition to the Updating state though, since the response contains the current status after processing the request it will have changed the state to Updating.

dcicutils/deployment_utils.py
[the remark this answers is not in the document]
netsettler: I often put `compute_` on the front of the name in that case just to underscore the difference. I don't know if that would be enough.

dcicutils/deployment_utils.py
```diff
@@ -66,15 +74,10 @@ def boolean_setting(settings, key, default=None):
 
 class EBDeployer:
 
+    # Options specifically related to our deployment
     S3_BUCKET = 'dcic-application-versions'
     EB_APPLICATION = '4dn-web'  # XXX: will need to change if this changes -Will
```
netsettler, line 79: Ha, it will sound funny to say it this way, but then make a constant of it.

I remember seeing such advice where someone was using 3.14159 in code and was told to make it be a constant in case it needs to change. Surprisingly, ordinary reasons to want to change the value of PI com up all the time, even other than [the stupid reasons](http://scienticity.net/wiki/Legislating_the_Value_of_Pi).

dcicutils/deployment_utils.py
```diff
@@ -319,15 +339,15 @@ def terminate_indexer_env(client, env_name):
             NOTE: there is 1 hr timeout before you can recreate one of these for the same application.
 
         :param client: boto3 elasticbeanstalk client
-        :param env_name:
+        :param env_name: one of: FF_ENV_INDEXER or CGAP_ENV_INDEXER
```
netsettler, line 342: Hmm. The idea of the function is not to use these variables at all. But it's a small matter so I'd leave it.

PR #87, opened 2020-05-18
Add a filtered_warnings context manager

netsettler:
Adds a 'filtered_warnings' context manager for disabling warnings in a small region of code when we already know about during system setup or testing.

### DIFF
--- a/dcicutils/misc_utils.py
+++ b/dcicutils/misc_utils.py
@@ -2,9 +2,13 @@
 This file contains functions that might be generally useful.
 """
 
+import contextlib
 import os
 import logging
+import warnings
 import webtest  # importing the library makes it easier to mock testing
+
+from typing import Type
 
 
 # Is this the right place for this? I feel like this should be done in an application, not a library.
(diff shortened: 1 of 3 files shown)


dcicutils/misc_utils.py
```diff
@@ -152,3 +156,20 @@ def get_setting_from_context(settings, ini_var, env_var=None, default=None):
         if env_var in os.environ:
             return os.environ.get(env_var)
     return settings.get(ini_var, default)
+
+
+# This should move to dcicutils.misc_utils for better sharing
```
willronchetti, line 161: This comment is now out of date 😄

netsettler: This comment about this being out of date is now out of date. :)

netsettler:
Yeah, this has two uses, of which oddly I had one of each case needed this weekend:
* In a test that calls old code (e.g., I have a test that compares legacy behavior to modern behavior), you can suppress the warning around the test.
* In the loadup sequence itself, there may be reasons we want to just mark the code as using something obsolete so that it doesn't get flagged on every test run as if we didn't know we had an issue. It's enough it has a "TO DO" but we have some versions pinned where we're tolerating old behavior because we have old versions and we will naturally confront these things as we upgrade. (If you disagree on this latter modality, and you might rightly do so, the time to flag that one is in the use, not the tool.)

PR #89, opened 2020-05-20
EB Template Verification

willronchetti:
- It is a common problem that EB will not recognize a template has been created if you try to `create_environment` immediately after creating a new template. Add a step to the indexer deploy that explicitly checks that the indexer template was successfully created.

### DIFF
--- a/dcicutils/deployment_utils.py
+++ b/dcicutils/deployment_utils.py
@@ -35,7 +35,7 @@ from git import Repo
 
 from dcicutils.env_utils import (
     get_standard_mirror_env, data_set_for_env, get_bucket_env, INDEXER_ENVS, is_fourfront_env, is_cgap_env,
-    FF_ENV_INDEXER, CGAP_ENV_INDEXER, is_indexer_env
+    FF_ENV_INDEXER, CGAP_ENV_INDEXER, is_indexer_env, indexer_env_for_env
 )
 from dcicutils.misc_utils import PRINT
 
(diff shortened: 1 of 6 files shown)


dcicutils/deployment_utils.py
```diff
@@ -265,19 +288,21 @@ def create_indexer_configuration_template(cls, env_name, size=None):
 
         # upload the template
         if is_cgap_env(env_name):
-            return eb_client.create_configuration_template(
+            eb_client.create_configuration_template(
                 ApplicationName=cls.EB_APPLICATION,
                 TemplateName=CGAP_ENV_INDEXER,
```
netsettler, line 293: Feels like we need a function indexer_env_for_environment(env) that returns either cgap_env_indexer or ff_env_indexer so that you can share these two bodies of code.

dcicutils/deployment_utils.py
```diff
@@ -227,6 +227,29 @@ def extract_env_name_configuration(cls, client, env_name):
...
+    def verify_template_creation(cls, client, template_name):
+        """ Does a get for the given template_name to verify EB has recognized that it is
+            available.
+
+        :param client: boto3 elasticbeanstalk client
+        :param template_name: name of template to check
+        :returns: True if template can be acquired
+        :raises: Exception if one is encountered
+        """
+        try:
```
netsettler, line 240: maybe this (my actual recommendation of these three):
```
    for retrying in (False, True):
        try:
            client.describe_configuration_settings(
                ApplicationName=cls.EB_APPLICATION,
                TemplateName=template_name
            )
        except Exception:
            if retrying:
                raise
            else:
                time.sleep(10)
    return True
```
Possibly also this, though it's more obscure:
```
    for attempt in range(2):
        if attempt != 0: time.sleep(10)
        try:
            client.describe_configuration_settings(
                ApplicationName=cls.EB_APPLICATION,
                TemplateName=template_name
            )
            return True
        except Exception:
            pass
    raise
```
and probably no one ever uses `else` in `for` but you can here and it's awful:
```
    for attempt in range(2):
        if attempt != 0: time.sleep(10)
        try:
            client.describe_configuration_settings(
                ApplicationName=cls.EB_APPLICATION,
                TemplateName=template_name
            )
            break
        except Exception:
            pass
    else:
        raise
    return True
```

willronchetti: Im taking the first one but flipping the retry order, feels more clear to me:
```
        for retrying in (True, False):
            try:
                client.describe_configuration_settings(
                    ApplicationName=cls.EB_APPLICATION,
                    TemplateName=template_name
                )
            except Exception:
                if not retrying:
                    raise
                else:
                    time.sleep(10)
        return True
```

PR #91, opened 2020-05-22
Add beanstalk_utils.compute_ff_stg_env and .compute_cgap_stg_env for foursight

netsettler:
This adds two functions to `dcicutils.beanstalk_utils` that are needed by a PR of Will's.

* `compute_ff_stg_env` (similar to existing `compute_ff_prd_env`, but returns the staging env)
* `compute_cgap_stg_env` (just returns `None` for now, but if `compute_cgap_prd_env` starts returning the value of CGAP_ENV_PRODUCTION_BLUE_NEW or CGAP_ENV_PRODUCTION_GREEN_NEW, this will start returning the other of those).

I used what I think is a more intuitive way to compute these, but took the implementation that was in the pending Foursight PR and used that as an additional test of correctness of my other technique.

Also, I had a branch open to improve unit tests in utils, so I repurposed that branch to contain these additions, hence the odd branch name and a number of opportunistic other tests that are riding along with this. This brings code coverage for `misc_utils` and `env_utils` to 100%.

### DIFF
--- a/test/test_misc_utils.py
+++ b/test/test_misc_utils.py
@@ -334,7 +334,7 @@ def test_virtual_app_crud_failure():
 
     simulated_error_message = "simulated error"
 
-    class FakeTestApp(VirtualAppError):
+    class FakeTestApp:
 
         def __init__(self, app, environ):
             ignored(app, environ)
@@ -371,28 +371,6 @@ def test_virtual_app_crud_failure():
                 assert str(e.raw_exception) == simulated_error_message
                 # assert isinstance(e.raw_exception, webtest.AppError)
                 assert isinstance(e.raw_exception, str)  # <-- TODO: This seems like a bug to me. -kmp 22-May-2020
-
-
-def test_virtual_app_post_failure():
-
-    class FakeTestApp(VirtualAppError):
-
-        def __init__(self, app, environ):
-            ignored(app, environ)
-
-        def get(self, url, **kwargs):
-            raise webtest.AppError("simulated error")
-
-    with mock.patch.object(VirtualApp, "HELPER_CLASS", FakeTestApp):
-
-        app = FakeApp()
-        environ = {'some': 'stuff'}
-
-        vapp = VirtualApp(app, environ)
-
-        with pytest.raises(VirtualAppError):
-            vapp.get("http://fixture.4dnucleome.org/some/url")
-
 
 
 def test_filtered_warnings():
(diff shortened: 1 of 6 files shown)


test/test_misc_utils.py
```diff
@@ -293,6 +294,107 @@ def test_virtual_app_patch_json():
...
+        ]
+
+        for operation in operations:
+            try:
+                operation()
+            except Exception as e:
+                assert isinstance(e, VirtualAppError)  # NOTE: not webtest.AppError, which is what was raised
+                assert str(e.raw_exception) == simulated_error_message
+                # assert isinstance(e.raw_exception, webtest.AppError)
+                assert isinstance(e.raw_exception, str)  # <-- TODO: This seems like a bug to me. -kmp 22-May-2020
```
willronchetti, line 373: `e` is cast to `str` when passed to VirtualAppError, see VirtualApp.

test/test_misc_utils.py
```diff
@@ -293,6 +294,107 @@ def test_virtual_app_patch_json():
...
+
+    # And the repr is the same.
+    assert repr(e2) == str(e2)
+
+
+def test_virtual_app_crud_failure():
+
+    simulated_error_message = "simulated error"
+
+    class FakeTestApp(VirtualAppError):
```
willronchetti, line 337: Why does this inherit from `VirtualAppError`?

test/test_misc_utils.py
```diff
@@ -293,6 +294,107 @@ def test_virtual_app_patch_json():
...
+            raise webtest.AppError("simulated error")
+
+    with mock.patch.object(VirtualApp, "HELPER_CLASS", FakeTestApp):
+
+        app = FakeApp()
+        environ = {'some': 'stuff'}
+
+        vapp = VirtualApp(app, environ)
+
+        with pytest.raises(VirtualAppError):
```
willronchetti, line 393: I'm not clear on why (syntactically?) having `FakeTestApp` inherit `VirtualAppError` causes the `raise` in `get` to be raised as a `VirtualAppError`. Something is missing in my understanding here.

test/test_misc_utils.py
netsettler: Yeah, I looked at that. I wasn't sure why, though. That's information-losing.

test/test_misc_utils.py
netsettler: As noted over Zoom, this was various errors conspiring -- really the parent class didn't matter, so I've removed that. The fact that there was an unrelated point made about this error class in a comment farther down was just coincidence. I think since we talked we're all set on this.

test/test_misc_utils.py
netsettler: OK, we discussed this on Zoom and agreed it shouldn't do that cast, so I've made that change.

test/test_misc_utils.py
netsettler: Same buggy issue now fixed. In the case of the second test, I just removed it as redundant. The crud test tests all three of `get`, `post_json`, and `patch_json`.

PR #92, opened 2020-05-22
Deployment options change in service of C4-166.

netsettler:
This adds a **DeployConfigManager** that can be used in Fourfront and CGAP to help resolve all the logic with **SKIP**, **STRICT**, and **WIPE_ES**. 

In service of that, I introduced `compute_prd_env_for_env` to do the obvious idiom of doing either `compute_ff_prd_env` or `compute_cgap_prd_env`.

Opportunistic other changes riding along:

* Renames `Deployer` to `IniFileManager`, though keeps the old name as an alias for compatibility for now.
* Removes some noisy logging in the internal support for `whodaman` and its more modern namings (`compute_ff_prd_env` and `compute_cgap_prd_env`, and the new `compute_prd_env_for_env`).

Some unit tests are owed.  This will remain a draft until that's fixed.

### DIFF
--- a/dcicutils/beanstalk_utils.py
+++ b/dcicutils/beanstalk_utils.py
@@ -234,9 +234,7 @@ def _compute_prd_env_for_project(project):
     magic_cname = CGAP_MAGIC_CNAME if project == 'cgap' else FF_MAGIC_CNAME
     client = boto3.client('elasticbeanstalk', region_name=REGION)
     res = describe_beanstalk_environments(client, ApplicationName="4dn-web")
-    # logger.info(res)
     for env in res['Environments']:
-        # logger.info(env)
         if env.get('CNAME') == magic_cname:
             # we found data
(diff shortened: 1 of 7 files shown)


dcicutils/beanstalk_utils.py
```diff
@@ -233,9 +234,9 @@ def _compute_prd_env_for_project(project):
     magic_cname = CGAP_MAGIC_CNAME if project == 'cgap' else FF_MAGIC_CNAME
     client = boto3.client('elasticbeanstalk', region_name=REGION)
     res = describe_beanstalk_environments(client, ApplicationName="4dn-web")
-    logger.info(res)
+    # logger.info(res)
```
willronchetti, line 237: I would just delete these lines altogether. They didn't serve much purpose before IMO and certainly don't now.

netsettler: Done.

PR #93, opened 2020-06-22
Auth env fix

sbreiff:
Fix in unified_authentication in ff_utils so that 'data' and 'staging' (as well as 'fourfront-green' and 'fourfront-blue') return correct environment names. Previously this relied on beanstalk_utils.BEANSTALK_PROD_BUCKET_ENVS but the bucket name doesn't match the environment name in the case of fourfront prod/staging environments.

### DIFF
--- a/dcicutils/env_utils.py
+++ b/dcicutils/env_utils.py
@@ -217,6 +217,17 @@ def prod_bucket_env(envname):
 
 
 def get_prd_or_stg_env(envname):
+    """
+    Given a production-class env label, returns the env name.
+    For other envnames that aren't production envs, this returns None.
+
+    The envname is something that is either a staging or production env, in particular something
+    that is_stg_or_prd_env returns True for.
+
+    The purpose is to return the envname when shorthand env labels like 'data' or 'staging' are
+    used in place of env names like fourfront-blue or fourfront-green. Should work for fourfront
+    as well as CGAP.
+    """
     if envname == 'data':
         use_env = bs_utils.compute_ff_prd_env()
     elif envname == 'staging':
--- a/dcicutils/env_utils.py
+++ b/dcicutils/env_utils.py
@@ -1,5 +1,6 @@
 import os
 from .misc_utils import get_setting_from_context
+from . import beanstalk_utils as bs_utils
 
 FF_ENV_DEV = 'fourfront-dev'  # Maybe not used
 FF_ENV_HOTSEAT = 'fourfront-hotseat'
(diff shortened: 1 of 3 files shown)


dcicutils/env_utils.py
```diff
@@ -215,6 +216,29 @@ def prod_bucket_env(envname):
     return BEANSTALK_PROD_BUCKET_ENVS.get(envname)
 
 
+def get_prd_or_stg_env(envname):
```
netsettler, line 219: There appear to be no tests associated with this PR. Minimally new functions call for new tests. But also the place that is changed to use this function also needed a test added showing what was broken prior to the change and fixed by the change, so that if further changes are needed it can be done without breaking something.

As an aside, a few weeks ago, I spent time getting env_utils.py (and several other sub-libraries of dcicutils) up to 100% code coverage--not that that proves we cover all the cases even now.  Code correctness depends on both control flow and the data available at each choice point, so it's possible for there to be situations not tested even when all the code paths have been taken. So I mostly just mean we need not to backslide on our commitment to good testing especially in a library like this that is so centrally important to several other big facilities working correctly.

The reason this came up is that it's possible this change is the cause of [C4-201](https://hms-dbmi.atlassian.net/browse/C4-201). It might not be, too.  But test cases might have made that easier to know.

sbreiff: You're right, tests would be helpful, I wasn't thinking of it at the time. I didn't want to leave problems right when I go on vacation so I created a branch with a couple tests for these. I probably won't be available to make changes, but I can send you the PR and you can make modifications if necessary.

It seems that unified_authentication was previously only being tested with integrated_ff i.e. mastertest, so I think the second test with other environments will be particularly helpful as this seemed to be where the problem had been arising. 

For the sake of record-keeping I'll post here the error I had been getting with unified_authentication (via generate-ontology in fourfront) before:

```
$ bin/generate-ontology --outfile src/encoded/ontology_terms_20200622.json --env data --pretty
Writing to src/encoded/ontology_terms_20200622.json
Traceback (most recent call last):
  File "bin/generate-ontology", line 81, in <module>
    sys.exit(encoded.commands.generate_ontology.main())
  File "/Users/sarah/fourfront/src/encoded/commands/generate_ontology.py", line 1018, in main
    connection = connect2server(args.env, key)
  File "/Users/sarah/fourfront/src/encoded/commands/generate_ontology.py", line 448, in connect2server
    key = unified_authentication(None, env)
  File "/Users/sarah/work/.virtualenvs/ff/lib/python3.6/site-packages/dcicutils/ff_utils.py", line 1092, in unified_authentication
    auth = s3_utils.s3Utils(env=use_env).get_access_keys()
  File "/Users/sarah/work/.virtualenvs/ff/lib/python3.6/site-packages/dcicutils/s3_utils.py", line 35, in __init__
    self.url = get_beanstalk_real_url(env)
  File "/Users/sarah/work/.virtualenvs/ff/lib/python3.6/site-packages/dcicutils/beanstalk_utils.py", line 319, in get_beanstalk_real_url
    bs_info = beanstalk_info(env)
  File "/Users/sarah/work/.virtualenvs/ff/lib/python3.6/site-packages/dcicutils/beanstalk_utils.py", line 288, in beanstalk_info
    return res['Environments'][0]
IndexError: list index out of range
```
Basically, it seemed using the env "data" had ultimately been pointing to "fourfront-webprod" as an env that doesn't exist anymore (it's a bucket but not an env).

PR #94, opened 2020-06-27
Tests auth env fix

sbreiff:
Tests added for PR#93.


test/test_ff_utils.py
```diff
@@ -156,6 +156,23 @@ def test_unified_authentication(integrated_ff):
     assert 'Must provide a valid authorization key or ff' in str(exec_info.value)
 
 
+def test_unified_authentication_more_envs():
+    key1 = ff_utils.unified_authentication(ff_env="data")
```
netsettler, line 160: I have a concern that this test does not test, especially for the 'staging' and 'data' frameworks, what the credential acquired is for the correct site. My intuition is that it needs to control for all the variabilities that are selecting either fourfront-green or fourfront-blue and make sure that the test succeeds not only in the moment it is run but would also succeed if the stars were aligned differently and the swap was the other way. Sometimes otherwise tests like this mysteriously start failing when the big switch is thrown (cname swap). So I think the items tested here are an improvement because they do test that something came back under some circumstances, but they aren't the full set of things we should be testing. I'm willing to look at how that might be done but not today. Since this is test-only stuff, it can presumably sit in a holding pattern for a few days until we've had a chance to work through this.

test/test_env_utils.py
```diff
@@ -66,6 +66,22 @@ def test_prod_bucket_env():
     assert prod_bucket_env('fourfront-cgapwolf') is None
 
 
+def test_get_prd_or_stg_env():
+    assert get_prd_or_stg_env('data') in ['fourfront-green', 'fourfront-blue']
+    assert get_prd_or_stg_env('staging') in ['fourfront-green', 'fourfront-blue']
```
netsettler, line 71: This should probably mock a `boto3` client that implements `describe_environments` in a way that allows these to be tested more precisely and then test both states of the `CNAME` so that we know that this is doing the predicted thing in both cases; that is, when the `CNAME` is set one way, one of these should return blue and the other green.

Even without doing that, I think it would be an improvement to do (untested):
```
fourfront_prod_envs = {'fourfront-green', 'fourfront-blue'}
prd = get_prd_or_stg_env('data')
assert prd in fourfront_prod_envs
assert get_prd_or_stg_env('staging') == (fourfront-prod_envs - {prd}).pop()
```

test/test_ff_utils.py
sbreiff: Well, keys are same for data and staging, so the end result should be the same, though the environment it has to look up is different. I made some edits that ensures it finds the same keys for data/staging/green/blue but as to whether the correct env is found, I assume we would need to test for that on whatever function actually looks that up. I haven't had time to look into whether those tests are there yet.

test/test_ff_utils.py
```diff
@@ -156,6 +156,23 @@ def test_unified_authentication(integrated_ff):
...
+    assert len(key2) == 2
+    key3 = ff_utils.unified_authentication(ff_env="fourfront-green")
+    assert len(key3) == 2
+    key4 = ff_utils.unified_authentication(ff_env="fourfront-blue")
+    assert len(key4) == 2
+    key5 = ff_utils.unified_authentication(ff_env="fourfront-cgap")
+    assert len(key5) == 2
+    with pytest.raises(Exception) as exec_info:
+        ff_utils.unified_authentication(ff_env="fourfront-data")
+        print(exec_info.value)
```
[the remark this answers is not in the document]
netsettler: This looks like it was taken care of. Thanks.

test/test_ff_utils.py
netsettler: I checked that s3Utils has code for getting the common prod bucket right now while these are shared buckets. We may need to change the implementation at some point, to make the credentials be different, but I think this test would notice that change and we'd end up having to write better tests at that time. So I think this is OK.

netsettler:
@sbreiff and @aschroed, [I added an additional test](https://github.com/4dn-dcic/utils/pull/94/commits/60618005eb68cb595b176aa602e462ec2706c914) that mocks out s3Utils and just checks that whatever credential is used and wherever it comes from (either from the function arguments to `ff_utils.unified_authorization` or from the S3 bucket that it grabs it from if it doesn't get it as an argument), it gets decoded correctly (to the best of my understanding of what correctly is). The reason I want this test is that I'll be suggesting a rewrite of that function, and I want something that is reasonably rigorous about noticing any bugs I might introduce. So you can also think of this added test as 'test first' for the thing I'm going to do later, but this shows it's compatible.

test/test_env_utils.py
netsettler: This is no longer relevant because you took out the function it applied to. But if you're curious about the mock stuff I was referring to, I was thinking of something similar to what I did in the mocks on [the added test in the last commit](https://github.com/4dn-dcic/utils/pull/94/commits/60618005eb68cb595b176aa602e462ec2706c914).

PR #97, opened 2020-07-11
Protect against CNAME swap during testing s3Utils (C4-225)

netsettler:
In order to address the problem described in [C4-225](https://hms-dbmi.atlassian.net/browse/C4-225), this fix moves the computation of the interesting value (the current production beanstalk) to right before it gets used, so that it won't go stale waiting to be used.  I got rid of the loop so that it wouldn't sit for the execution of the first pass through the loop before needing the second value. Instead, I just turned the loop body into a function, and each of the two calls it was going to iterate over into visibly distinct calls to that function so that the second value is computed as late as can be.

**Testing**

Probably the conditions of the failure are hard to make happen, although we _could_ arrange to run tests in a coordinated way with the CNAME swap sometime to be sure. Still, it's hard to see how doing it this way can make anything worse so it seems worth making this change regardless.

Also, this adds a call to `date` to the `.travis.yml` so it's easier to quickly tell what time the test ran in case of future problems.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ jobs:
     - python: "3.6"
     - python: "3.7"
 before_install:
+  - date
   - pip install poetry
 install:
   - echo "Installing through poetry then running tests"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "dcicutils"
-version = "0.33.1"
+version = "0.33.2"
 description = "Utility package for interacting with the 4DN Data Portal and other 4DN resources"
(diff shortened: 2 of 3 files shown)


netsettler:
> I see cgap is still parameterized? Any concerns there?

As you note, probably not. But your comment is still a good one so I made the change so that we're better prepared when there is a split.

> And we aren't worrying about other fourfront envs other than data and staging?

I hadn't noticed that untilI made this change.  It was sort of outside the scope of what I was fixing but it's really conspicuous once seeing it.  I've uploaded a fix for that as well.  If it all passes, I'll go ahead with the merge unless you see some problem.

PR #98, opened 2020-07-16
Add LockoutManager and RateManager

netsettler:
Adds a class `RateManager` which does rate-limiting (e.g., to accomplish lockout for a threshold period of time on a guarded operation, or to avoid throttling). This is a generalization of the original `LockoutManager` (which only managed the case of 1 call per interval time).

This does not do synchronization locking in case of race conditions. I think that's still an improvement over what we are doing now, but it might be a useful addition

Note that `LockoutManager` (factored out of snovault's purge code) is included, but only as a comment. Per @willronchetti's comments, `RateManager` should be enough, and so I've eliminated the redundancy. Since there was some investment in producing the other, I've kept it as a comment for now in case we need to revert to or refer to it. But we can make the commented-out code go away soon when `RateManager` is fully installed and running smoothly.

### DIFF
--- a/dcicutils/misc_utils.py
+++ b/dcicutils/misc_utils.py
@@ -546,7 +546,11 @@ class RateManager:
         expiration_delta = datetime.timedelta(seconds=self.interval_seconds)
         latest_expiration = now + expiration_delta
         soonest_expiration = latest_expiration
-        soonest_expiration_pos = None
+        # This initial value of soonest_expiration_pos is arbitrarily chosen, but it will normally be superseded.
+        # The only case where it's not overridden is where there were no better values than the latest_expiration,
+        # so if we wait that amount, all of the slots will be ready to be reused and we might as well use 0 as any.
+        # -kmp 19-Jul-2020
+        soonest_expiration_pos = 0
         for i, expiration_time in enumerate(self.timestamps):
             if expiration_time <= now:  # This slot was unused or has expired
                 self.timestamps[i] = latest_expiration
@@ -560,5 +564,5 @@ class RateManager:
                 self.wait_hook(wait_seconds=sleep_time_needed, next_expiration=soonest_expiration)
             self.log.warning("Waiting %s seconds before attempting %s." % (sleep_time_needed, self.action))
             time.sleep(sleep_time_needed)
-        # It will have expired now, so grab that slot
+        # It will have expired now, so grab that slot. We have to recompute the 'now' time because we slept in between.
         self.timestamps[soonest_expiration_pos] = datetime.datetime.now() + expiration_delta
--- a/dcicutils/misc_utils.py
+++ b/dcicutils/misc_utils.py
@@ -359,108 +359,112 @@ def utc_today_str():
     return datetime.datetime.strftime(datetime.datetime.utcnow(), "%Y-%m-%d")
 
 
-class LockoutManager:
-    """
-    This class is used as a guard of a critical operation that can only be called within a certain frequency.
-    e.g.,
-
-        class Foo:
-            def __init__(self):
-                # 60 seconds required between calls, with a 1 second margin of error (overhead, clocks varying, etc)
-                self.lockout_manager = LockoutManager(action="foo", lockout_seconds=1, safety_seconds=1)
-            def foo():
-                self.lockout_manager.wait_if_needed()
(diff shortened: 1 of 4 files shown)


dcicutils/misc_utils.py
```diff
@@ -325,3 +325,100 @@ def apply_dict_overrides(dictionary: dict, **overrides) -> dict:
 
 def utc_today_str():
     return datetime.datetime.strftime(datetime.datetime.utcnow(), "%Y-%m-%d")
+
+
+class LockoutManager:
```
willronchetti, line 330: It would be nice to have some more documentation here on how this works/is used.

dcicutils/misc_utils.py
```diff
@@ -325,3 +325,100 @@ def apply_dict_overrides(dictionary: dict, **overrides) -> dict:
...
+
+    def update_timestamp(self):
+        """
+        Explicitly sets the reference time point for computation of our lockout.
+        This is called implicitly by .wait_if_needed(), and for some situations that may be sufficient.
+        """
+        self._timestamp = datetime.datetime.now()
+
+
+class RateManager:
```
willronchetti, line 382: Same documentation comment here.

dcicutils/misc_utils.py
```diff
@@ -325,3 +325,100 @@ def apply_dict_overrides(dictionary: dict, **overrides) -> dict:
...
+        # never exceed self.effective_lockout_seconds, so
+        #   0 <= wait_seconds <= self.effective_lockout_seconds
+        wait_seconds = max(0.0, self.effective_lockout_seconds - seconds_since_last_purge)
+        if wait_seconds > 0.0:
+            shared_message = ("Last %s attempt was at %s (%s seconds ago)."
+                              % (self.action, self._timestamp, seconds_since_last_purge))
+            if self.lockout_enabled:
+                action_message = "Waiting %s seconds before attempting another." % wait_seconds
+                if self.log:
+                    self.log.warning("%s %s" % (shared_message, action_message))
```
willronchetti, line 366: I notice you do `if self.log` in some places here, wouldn't it be better to just have a helper method like `log_if_you_can`? Moreover, it feels like these two classes are very similar - can they be refactored to extend a base class? Is it worthwhile to do so?

netsettler: Theoretically one can be a special case of the other, where its `allowed_attempts=1` and the length of the interval is similar to the lockout time. The thing is that the n=1 case has an ability to ask about its timestamp in a way that you _could_ generalize to the multiple case, but it's tricky. On the other hand, I will do the experiment of just using a RateManager on the purge code and see if that makes any difference. The methods like timestamp mostly affect how we unit-test things. Probably if the unit tests were rewritten, it wouldn't matter. After all, the RateManager has its own unit tests.

netsettler: As for the `.log` that's slightly different. The purpose is to be a proxy for the other log. Theoretically, as a library, you don't know how logging is accomplished and you can't just find the logger because that's a lexical matter. So someone does a 
```
LOG = some_log_factory.make_some_log(__file__)
```
kind of call that sets up a log specific to this file and specific to its needs, and while you can make a logger that logs _somewhere_, it's not possible to guess the right place.

That's a separate question from whether I should have done:
```
def warn(*args, **kwargs):
    if self.log:
        self.log.warning(*args, **kwargs)
```
or
```
import logging
...
class LockoutManager:
    DEFAULT_LOGGER = logging
    def __init__(..., log=None):
        self.log = log or DEFAULT_LOGGER
```
The latter seems worth doing, I think. That would give the default effect of your `log_if_you_can` by making sure you always can, and then one could use `self.log.warning(...)` without conditionals.

dcicutils/misc_utils.py
netsettler: Doc added.

dcicutils/misc_utils.py
netsettler: Doc added.

netsettler:
The only material change I've made here is to comment out `LockManager`, which is responsive to code review. I'm going to go ahead and merge this. Not only are tests passing here, but I built a version of `snovault` with this and its tests pass, and I built CGAP and Fourfront with it, and those tests pass as well.

PR #99, opened 2020-07-17
unzip s3 to s3: option not to store data

SooLee:


SooLee:
I added an option to not store/return the entire zip content in unzip_s3_to_s3 (the function does s3 to s3 stream with or without this option, sometimes we don't need to return the content for further use) - I need this change to reduce memory for bamsnap.

dcicutils/s3_utils.py
```diff
@@ -220,8 +225,9 @@ def unzip_s3_to_s3(self, zipped_s3key, dest_dir, acl=None):
                 # files twice that we want to further interogate
                 the_file = zipstream.open(file_name, 'r').read()
                 file_to_find = file_name.split('/')[-1]
-                ret_files[file_to_find] = {'s3key': s3_key,
-                                           'data': the_file}
+                if not do_not_store:
+                    ret_files[file_to_find] = {'s3key': s3_key,
+                                               'data': the_file}
                 self.s3_put(the_file, s3_file_name, acl=acl)
```
netsettler, line 231: Shouldn't this be indented into the `if`? It looks like if you set this variable it just lies about what it did.

dcicutils/s3_utils.py
```diff
@@ -187,7 +187,12 @@ def read_s3_zipfile(self, s3key, files_to_extract):
                 ret_files[name] = zipstream.open(zipped_filename).read()
         return ret_files
 
-    def unzip_s3_to_s3(self, zipped_s3key, dest_dir, acl=None):
+    def unzip_s3_to_s3(self, zipped_s3key, dest_dir, acl=None, do_not_store=False):
```
netsettler, line 190: Why isn't this `store_result=True`? It hurts my head to be looking at double negatives.

test/test_s3_utils.py
```diff
@@ -220,3 +220,23 @@ def test_unzip_s3_to_s3_2(integrated_s3_info):
...
+    filename = integrated_s3_info['zip_filename']
+    integrated_s3_info['s3Obj'].s3_delete_dir(prefix)
+
+    # ensure this thing was deleted
+    # if no files there will be no Contents in response
+    objs = integrated_s3_info['s3Obj'].s3_read_dir(prefix)
+    assert [] == objs.get('Contents', [])
+
+    # now copy to that dir we just deleted
+    ret_files = integrated_s3_info['s3Obj'].unzip_s3_to_s3(filename, prefix, do_not_store=True)
```
netsettler, line 238: This isn't testing whehter the file got written, only what the result is.

SooLee: The file writing is tested in the lines below. This is the same test repeated three times with slightly different contexts (you'd be able to see it if you open the whole file)

dcicutils/s3_utils.py
SooLee: What do you mean? Yes, it is indented into the `if`.

dcicutils/s3_utils.py
SooLee: Yeah I know, either way works but since the default is store and we have used this function for years assuming it will store the data, I wanted to make it explicit that we do not want to store it.

netsettler: I think a default of `store=True` still makes it explicit in that
* unzip_s3_to_s3(zipped_key, dir, acl=acl) is the same as always, just implying store=True
* unzip_s3_to_s3(zipped_key, dir, acl=acl, store=False) is explicit about not storing.

Although really a name that said what is not being stored might be helpful, regardless of what the boolean sense of the flag is.

It sounds weird to say, but It feels like if this argument is given you're saying "take the time to do the operation, but don't put the output anywhere". Other than just consuming time, I don't see what is going on in that case. It might be doing something and I might not be seeing it, so please do feel free to point to what it doees other than that one 'put' that is either being suppressed or not. Surely the argument is not only about suppressing the return value, since it doesn't say "do_not_tell_me_what_you_did" but instead says "do_not_do_something".

Sorry if I seem to be repeating myself 6 different ways, but I'm trying to offer you different ways to see my confusion and fix me. :)

dcicutils/s3_utils.py
```diff
@@ -220,8 +225,9 @@ def unzip_s3_to_s3(self, zipped_s3key, dest_dir, acl=None):
                 # files twice that we want to further interogate
                 the_file = zipstream.open(file_name, 'r').read()
                 file_to_find = file_name.split('/')[-1]
```
netsettler, line 227: This line is equivalent to, and better written as:
```
file_to_find = os.path.basename(file_name)
```
As an example:
```
>>> os.path.basename("foo/bar/baz.html")
'baz.html'
>>> "foo/bar/baz.html".split("/")[-1]
'baz.html'
```
HOWEVER, I don't think the value of this variable is used so I'm not sure why the line is here at all.

test/test_s3_utils.py
```diff
@@ -220,3 +220,23 @@ def test_unzip_s3_to_s3_2(integrated_s3_info):
 
     objs = integrated_s3_info['s3Obj'].s3_read_dir(prefix)
     assert objs.get('Contents', None)
+
+
+def test_unzip_s3_to_s3(integrated_s3_info):
```
netsettler, line 225: This test name duplicates an earlier name. I probably `test_unzip_s3_to_s3_do_not_store` or something like that would help?

netsettler:
My PR with some changes to this is [utils PR #100](https://github.com/4dn-dcic/utils/pull/100).

PR #104, opened 2020-07-28
Make LockoutManager available after all

netsettler:
We were thinking `LockManager` was redundant if we had `RateManage`r, but I suspect `RateManager` is causing some timing errors. It is slightly more complicated and so could have a bug that's harder to spot. For that reason, I want to go ahead and carry the `LockManager` abstraction, too.

### DIFF
--- a/dcicutils/misc_utils.py
+++ b/dcicutils/misc_utils.py
@@ -216,7 +216,8 @@ class Retry:
             self.wait_multiplier = wait_multiplier
             self.wait_adjustor = self.make_wait_adjustor(wait_increment=wait_increment, wait_multiplier=wait_multiplier)
 
-        def make_wait_adjustor(self, wait_increment=None, wait_multiplier=None):
+        @staticmethod
+        def make_wait_adjustor(wait_increment=None, wait_multiplier=None):
             """
             Returns a function that can be called to adjust wait_seconds based on wait_increment or wait_multiplier
             before doing a retry at each step.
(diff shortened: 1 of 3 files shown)


dcicutils/misc_utils.py
```diff
@@ -368,112 +369,108 @@ def utc_today_str():
...
+        to comply with rate-limiting declared in the creation of this LockoutManager.
+
+        NOTE WELL: It is presumed that all calls are coming from this source. This doesn't have ESP that would
+        detect or otherwise accommodate externally generated calls, so violations of rate-limiting can still
+        happen that way. This should be sufficient for sequential testing, and better than nothing for
+        production operation.  This is not a substitute for responding to server-initiated throttling protocols.
+        """
+        now = datetime.datetime.now()
+        # Note that this quantity is always positive because now is always bigger than the timestamp.
+        seconds_since_last_purge = (now - self._timestamp).total_seconds()
```
willronchetti, line 451: This variable name looks specific to a particular use case.

dcicutils/misc_utils.py
```diff
@@ -368,112 +369,108 @@ def utc_today_str():
...
+        action str: A noun or noun phrase describing the action being guarded.
+        enabled bool: A boolean controlling whether this facility is enabled. If False, waiting is disabled.
+        log object: A logger object (supporting operations like .debug, .info, .warning, and .error).
+        """
+
+        # This makes it easy to turn off the feature
+        self.lockout_enabled = enabled
+        self.lockout_seconds = lockout_seconds
+        self.safety_seconds = safety_seconds
+        self.action = action
```
willronchetti, line 419: Would you set action to the name of the operation being guarded? Is this just for debugging?

dcicutils/misc_utils.py
netsettler: No, it's just a matter of making it debuggable. If time keeps changing out from underneath you while you're in the debugger, it's hard to step. It's not any different than doing `foo = computation()` where computation could change due to external factors and any attempt to put the result in a variable inevitably freezes the item for view at the risk that you're not picking up changes. But in return you get consistency.

dcicutils/misc_utils.py
netsettler: It's for error messages. It has no computational effect. (You might also have several of these and so you need a visual way to know which one you're talking to.)

dcicutils/misc_utils.py
netsettler: Oh, I see. The purge variable. RIght. Yeah, I'll fix that.

dcicutils/misc_utils.py
netsettler: I think the doc string in the `__init__` method is pretty clear: "action str: A noun or noun phrase describing the action being guarded."

dcicutils/misc_utils.py
netsettler: Changing it to `seconds_since_last_attempt`.

PR #105, opened 2020-07-29
C4-245 Multi-Get API

willronchetti:
- Allow authenticated users to use the `get_bulk_uuids_embedded` function to get the embedded view of many uuids (of the same item type from a single index). The idea is for this method to replace `expand_es_metadata` where possible as it seems to perform much better.

### DIFF
--- a/dcicutils/es_utils.py
+++ b/dcicutils/es_utils.py
@@ -104,15 +104,18 @@ def get_bulk_uuids_embedded(client, index, uuids, is_generator=False):
 
     :returns: list of embedded views of the given uuids, if any
     """
+    def return_generator(resp):
+        for d in resp['docs']:
+            yield d['_source']['embedded']
+
     final_result = []
     response = client.mget(body={  # XXX: this could still be slow even if you use is_generator
         'docs': [{'_id': _id,
                   'source': ['embedded.*'],
                   '_index': index} for _id in uuids]
         })
-    if is_generator:
-        for doc in response['docs']:
-            yield doc['_source']['embedded']
+    if is_generator is True:
+        return return_generator(response)
     else:
         for doc in response['docs']:
             final_result.append(doc['_source']['embedded'])
--- a/dcicutils/es_utils.py
+++ b/dcicutils/es_utils.py
@@ -88,7 +88,7 @@ def execute_lucene_query_on_es(client, index, query):
         return None
 
 
-def get_bulk_uuids_embedded(client, index, uuids):
+def get_bulk_uuids_embedded(client, index, uuids, is_generator=False):
     """
     Gets the embedded view for all uuids in an index with a single multi-get ES request.
 
@@ -100,18 +100,20 @@ def get_bulk_uuids_embedded(client, index, uuids):
     :param client: elasticsearch client
     :param index: index to search
     :param uuids: list of uuids (all of the same type)
+    :param is_generator: whether to use a generator over the response (NOT paginate)
 
     :returns: list of embedded views of the given uuids, if any
     """
     final_result = []
-    try:
-        response = client.mget(body={
-            'docs': [{'_id': _id,
-                      'source': ['embedded.*'],
-                      '_index': index} for _id in uuids]
+    response = client.mget(body={  # XXX: this could still be slow even if you use is_generator
+        'docs': [{'_id': _id,
+                  'source': ['embedded.*'],
+                  '_index': index} for _id in uuids]
         })
+    if is_generator:
+        for doc in response['docs']:
+            yield doc['_source']['embedded']
+    else:
         for doc in response['docs']:
             final_result.append(doc['_source']['embedded'])
-    except Exception as e:
-        raise e
-    return final_result
+        return final_result
--- a/dcicutils/es_utils.py
+++ b/dcicutils/es_utils.py
@@ -86,3 +86,32 @@ def execute_lucene_query_on_es(client, index, query):
     except KeyError:
         logger.error('Searching index %s with query %s gave no results' % (index, query))
         return None
+
+
+def get_bulk_uuids_embedded(client, index, uuids):
+    """
+    Gets the embedded view for all uuids in an index with a single multi-get ES request.
+
+    NOTE: because an index is required, when passing uuids to this method they must all be
+    of the same item type. The index can be determined by:
+            ''.join([eb_env_name, item_type])
+            ex: fourfront-mastertestuser or fourfront-mastertestfile_format
+
+    :param client: elasticsearch client
+    :param index: index to search
+    :param uuids: list of uuids (all of the same type)
+
+    :returns: list of embedded views of the given uuids, if any
+    """
+    final_result = []
+    try:
+        response = client.mget(body={
+            'docs': [{'_id': _id,
+                      'source': ['embedded.*'],
+                      '_index': index} for _id in uuids]
+        })
+        for doc in response['docs']:
+            final_result.append(doc['_source']['embedded'])
+    except Exception as e:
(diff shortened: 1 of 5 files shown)


dcicutils/qa_utils.py
```diff
@@ -4,13 +4,27 @@
...
 from .misc_utils import PRINT, ignored, Retry
 
 
+@contextlib.contextmanager
+def timed():
+    """ A simple context manager that will time how long it spends in context. Useful for debugging. """
+    start = time.time()
+    try:
+        yield
+    except Exception:
```
netsettler, line 21: Is this doing something? It seems like a no-op. Maybe you sometimes put code in the middle, in which case I'd write this (using `dcicutils.misc_utils.ignored` to keep tools from fussing about sometimes unused `e`...
```
except Exception as e:
    ignored(e)
    # Debugging code might want to go here.
    raise
```

dcicutils/qa_utils.py
```diff
@@ -4,13 +4,27 @@
...
+import time
 import io
 import os
 import pytz
 
 from .misc_utils import PRINT, ignored, Retry
 
 
+@contextlib.contextmanager
+def timed():
```
netsettler, line 16: I might either make this a class so the report can be customized by different methods or else make something that shows the info be an outboard function that becomes the default. e.g,
```
def show_elapsed_time(start, end):
    print('Elapsed: %s' % (end - start))
@contextlib.contextmanager
def timed(reporter=None):
    if reporter is None:
        reporter = show_elapsed_time
    ...
    finally:
        end = time.time()
        show_elapsed_time(end, start)
```

Alternatively the class-based way of doing it would just have that be a method call instead of a functional parameter.

The reason I care, though, is that I'm not sure print is always what you want. You might actually want to have it do logging of some kind, including possibly using that `debuglog` capability I just recently put in `dcicutils.miscutils`. But having it come out on stdout is sometimes disruptive.

test/test_es_utils.py
```diff
@@ -26,4 +27,34 @@ def test_lucene_query_basic(es_client_fixture):
...
\ No newline at end of file
+    assert len(results) == 1
+
+
+@pytest.mark.integrated
+def test_get_bulk_uuids_embedded(es_client_fixture):
+    """ Tests getting some bulk uuids acquired from search. """
+    uuids = ['1a12362f-4eb6-4a9c-8173-776667226988']  # only one uuid first
+    result = get_bulk_uuids_embedded(es_client_fixture, 'fourfront-mastertestuser', uuids)
+    assert len(result) == 1
```
netsettler, line 38: For each of these, I would perhaps minimally also do something to test that the set of results has the indicated uuids in it. But also and perhaps more important, if there is a way to test that it's getting the embedded view and not some other view, that would be good, perhaps by checking that a certain key was included that is part of the embedded view and that some other is excluded that is part of a more expansive view.

test/test_es_utils.py
```diff
@@ -26,4 +27,34 @@ def test_lucene_query_basic(es_client_fixture):
...
+    """ Tests that the new method out performs the new one. """
+    users = execute_lucene_query_on_es(client=es_client_fixture, index='fourfront-mastertestuser', query={})
+    uuids = [doc['_id'] for doc in users]
+
+    # this extracts the 10 desired uuids in embedded view using mget in 200 ms
+    with timed():
+        get_bulk_uuids_embedded(es_client_fixture, 'fourfront-mastertestuser', uuids)
+
+    # this gets 19 total uuids in 1700ms (~850 ms adjusted, so roughly 4x slower)
+    with timed():
```
netsettler, line 58: If you had the parameter I suggested above, you could get the timing info to use in the comparison and maybe require that the speed is better in one than the other...?

dcicutils/es_utils.py
```diff
@@ -86,3 +86,32 @@ def execute_lucene_query_on_es(client, index, query):
     except KeyError:
         logger.error('Searching index %s with query %s gave no results' % (index, query))
         return None
+
+
+def get_bulk_uuids_embedded(client, index, uuids):
```
netsettler, line 91: I'll have to think more on this, but I'll just ask the question aloud anyway: Are there security issues in accessing in bulk, since the individual items will have differing security characteristics?  Is that already handled in `client.mget` and not relevant here?

Also, if a problem results in getting one of these but not another, do you get a partial result? no result? a list of errors? ... Maybe again that is really an issue of mget, which I'll have to look into. I'm just thinking aloud here.

willronchetti: Since this API is exposed through dcicutils and requires an ES client, only authenticated users can do this operation. It isn't mean to be an externally facing (public) API. If we want to provide this ability application side, there would need to be greater protections in place.

test/test_es_utils.py
```diff
@@ -36,25 +37,37 @@ def test_get_bulk_uuids_embedded(es_client_fixture):
...
 
 @pytest.mark.integrated
 def test_get_bulk_uuids_outperforms_expand_es_metadata(integrated_ff, es_client_fixture):
     """ Tests that the new method out performs the new one. """
     users = execute_lucene_query_on_es(client=es_client_fixture, index='fourfront-mastertestuser', query={})
     uuids = [doc['_id'] for doc in users]
+    current_time = [None]  # pass by reference
+
+    def set_current_time(start, end, t=current_time):  # noqa I want this default argument to be mutable
+        t[0] = end - start
```
netsettler:
```
times = []
def set_current_time(start, end):
    times.append(end-start)
```
Then you don't have to do any copying between, can just call it twice, and can just make the second assertion do 
```
assert times[1] < times[0] / 2, "New functionality isn't twice as fast as clumsier technology."
```

dcicutils/qa_utils.py
```diff
@@ -12,17 +12,30 @@
 from .misc_utils import PRINT, ignored, Retry
 
 
+def show_elapsed_time(start, end):
+    """ Helper method for below that is the default - just prints the elapsed time. """
+    print('Elapsed: %s' % (end - start))
```
netsettler, line 17: This should use `PRINT` from the above import so that grepping for stray `print` won't find this. The purpose of the uppercase version is to say "this is intended programmatic output part of a published feature", not debugging. Mind you, this published feature is for debugging. But on the metalevel at which things matter, this intends to print.

aschroed:
I don't think this can necessarily replace expand_es_metadata as that function essentially iteratively goes through items - retrieves uuids of embedded items and goes through those repeating until it gets all linked items - expand_es_metadata is not written recursively per se but achieves the same goal.  So I think the new function replaces get_es_metadata unless I am missing something?

dcicutils/es_utils.py
```diff
@@ -86,3 +86,32 @@ def execute_lucene_query_on_es(client, index, query):
...
+    :returns: list of embedded views of the given uuids, if any
+    """
+    final_result = []
+    try:
+        response = client.mget(body={
+            'docs': [{'_id': _id,
+                      'source': ['embedded.*'],
+                      '_index': index} for _id in uuids]
+        })
+        for doc in response['docs']:
```
aschroed, line 113: Any concern about huge list?  Might set up as a generator and have a wrapper function to return a list with the option to use the generator directly if the result is huge?  similar to how search_metadata is set up.

netsettler: To @aschroed's comment...

(1) Is it possible that if it works this way, and its stack is made explicit, it could return the info retrieved in the time allotted _and_ a representation of partial state to resume from in order to continue?

(2) Is it possible that it could return json-lines format streaming and that this would get around the issue of a large list _and_ timeouts?

PR #108, opened 2020-08-14
Fix integrated ES setup (C4-275)

netsettler:
This contains a fix for C4-275, which was due to a wired notion of where the ES is for `fourfront-mastertest`. Now it will be discovered by looking at its health page.

I also fixed another issue that was probably broken only for me, where if you already have certain env variables set in your local environment, a test that supposedly checked what happens when you don't initialize failed because it assumed the variables wouldn't be set by default.  Better to set up the appropriate environment.

And I added more complete test coverage for s3_utils, bringing it up to 100%

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -1,15 +1,44 @@
 # flake8: noqa
+import boto3
 import pytest
 import os
+import requests
+
+from dcicutils.misc_utils import check_true
 from dcicutils.s3_utils import s3Utils
 from dcicutils.ff_utils import authorized_request
+from dcicutils.beanstalk_utils import describe_beanstalk_environments, REGION
+
+
+def _discover_es_info(envname):
+    try:
+        eb_client = boto3.client('elasticbeanstalk', region_name=REGION)
+        # Calling describe_beanstalk_environments is pretty much the same as doing eb_client.describe_environments(...)
+        # except it's robust against AWS throttling us for calling it too often.
+        envs = describe_beanstalk_environments(eb_client, EnvironmentNames=[envname])['Environments']
+        for env in envs:
+            if env.get('EnvironmentName') == envname:
+                cname = env.get('CNAME')
+                # TODO: It would be nice if we were using https: for everything. -kmp 14-Aug-2020
+                res = requests.get("http://%s/health?format=json" % cname)
+                health_json = res.json()
+                return health_json['elasticsearch'], health_json['namespace']
+    except Exception as e:
+        raise RuntimeError("Unable to discover elasticsearch info for %s:\n%s: %s" % (envname, e.__class__.__name__, e))
+
 
 # XXX: Refactor to config
 INTEGRATED_ENV = 'fourfront-mastertest'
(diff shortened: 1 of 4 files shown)


test/conftest.py
```diff
@@ -1,16 +1,45 @@
...
 import os
+import requests
+
+from dcicutils.misc_utils import check_true
 from dcicutils.s3_utils import s3Utils
 from dcicutils.ff_utils import authorized_request
+from dcicutils.beanstalk_utils import describe_beanstalk_environments, REGION
+
+
+def _discover_es_info(envname):
```
willronchetti, line 13: This utility is not needed - use `ff_utils.get_health_page`.

test/test_jh_utils.py
```diff
@@ -15,10 +17,19 @@ def initialize_jh_env(server):
 
 
 def test_import_fails_without_initialization():
-    # this fails because proper env variables were not set up
-    with pytest.raises(Exception) as exec_info:
-        from dcicutils import jh_utils  # NOQA
-    assert 'ERROR USING JUPYTERHUB_UTILS!' in str(exec_info.value)
+
+    # Loading this library fails if proper env variables were not set up.
+    # TODO: I'm not sure I think that's a good idea. Functions should fail, imports should not. -kmp 14-Aug-2020
```
willronchetti, line 22: Imports are functions to me 😄   - IMO imports should fail if on import the module determines it will not work, as is the case with this particular module. I would rather fail on import than fail mysteriously in a function that I shouldn't have been able to call in the first place.

test/conftest.py
netsettler: I was trying to avoid using that because we're testing that. I did a similar thing with computing blue/green. My thought was that the real "strength" of a wired name, even though it has weaknesses, is that it specifies an anchor that other tests will judge consistency against. If it uses those other tools, it's at risk of them being wrong. And maybe that argues for failing on the throttling or using a more low-tech other throttling. It might make sense to call that other function and compare the result to make sure we agree we're talking to the right page. It seems like a hard thing for the get_health_page thing to get wrong, but I'm ignoring that and thinking more in terms of what are the consequences if it got the health page for the wrong server, and it seems like the consequence is that a lot of integration tests are done to the wrong resource, which could put that resource at some risk.

In general, I think testing code is not like regular code in that sometimes things are inelegant, redundant, differing in syntax and that can be good because otherwise if you're too regular you may not trip up some subtle difference. The core value in algebra as a concept, whether high school algebra, abstract algebra, or the kind of weird programming algebra we do is that no matter which way you go to get an answer, you want to get the same answer. But in programming that's an ideal more than a reality, so testing goodness by going different routes can sometimes be a virtue in itself because there is additional confidence in something tested several ways.

test/test_jh_utils.py
netsettler: I get what you're saying. But then again it makes it hard to unit test. It means that static code analysis tools trip over it, thinking perhaps that the entire file is unreachable. The model I'm suggesting is that there are more reasons to load a library than to execute it.

test/conftest.py
netsettler: I made it do a consistency check as part of the setup. I'm going to leave it at that. :)

PR #109, opened 2020-08-18
Additions to env_utils and command_utils, and fixes for C4-278 and C4-279

netsettler:
Various changes, mostly in service of `SubmitCGAP`:

* New function `command_utils.yes_or_no` for use by `SubmitCGAP`.
* New functions `full_env_name`, `full_cgap_env_name`, and `full_fourfront_env_name` in `env_utils`.
* Fix for [file naming bug in dcicutils.data_utils (C4-278)](https://hms-dbmi.atlassian.net/browse/C4-278).
* Fix for [python 3.5 incompatibility for dcicutils v0.39.0 (C4-279)](https://hms-dbmi.atlassian.net/browse/C4-279).

Opportunistic:

* Disable `beanstalk_utils.create_bs` in staging & production. (There was a comment in the doc string saying not to do it, but it seemed prudent to add enforcement.)

### DIFF
--- a/dcicutils/beanstalk_utils.py
+++ b/dcicutils/beanstalk_utils.py
@@ -668,9 +668,12 @@ def create_bs(envname, load_prod, db_endpoint, es_url, for_indexing=False):
     Returns:
         dict: boto3 res from create_environment/update_environment
     """
+    if is_stg_or_prd_env(envname):
+        raise RuntimeError("beanstalk_utils.create_bs is not approved for production use.")
+
     client = boto3.client('elasticbeanstalk', region_name=REGION)
 
-    # deterimine the configuration template for Elasticbeanstal
+    # determine the configuration template for Elasticbeanstalk
     template = 'fourfront-base'
     if for_indexing:
         template = 'fourfront-indexing'
@@ -688,6 +691,7 @@ def create_bs(envname, load_prod, db_endpoint, es_url, for_indexing=False):
     # logic for mirrorEsEnv, which is used to coordinate elasticsearch
     # changes between fourfront data and staging
     if 'fourfront-webprod' in envname:
+        # TODO: This code is obsolete and needs to be upgraded. For now, the use of this on production is disabled.
         other_env = 'fourfront-webprod2' if envname == 'fourfront-webprod' else 'fourfront-webprod'
         mirror_es = get_es_build_status(other_env, max_tries=3)
         if mirror_es:
--- a/dcicutils/beanstalk_utils.py
+++ b/dcicutils/beanstalk_utils.py
@@ -668,6 +668,7 @@ def create_bs(envname, load_prod, db_endpoint, es_url, for_indexing=False):
     Returns:
         dict: boto3 res from create_environment/update_environment
     """
+    # TODO (C4-280): Reconsider this and other functionality.
     if is_stg_or_prd_env(envname):
         raise RuntimeError("beanstalk_utils.create_bs is not approved for production use.")
 
--- a/dcicutils/command_utils.py
+++ b/dcicutils/command_utils.py
@@ -1,0 +1,38 @@
+from .misc_utils import PRINT
+
+
+def yes_or_no(question, quick=None, default=None):
+    """
+    Loops asking a question interactively until it gets a 'yes' or 'no' response. Returns True or False accordingly.
+
+    :param question: The question to ask (without prompts for possible responses, which will be added automatically).
+    :param quick: Whether to allow short-form responses.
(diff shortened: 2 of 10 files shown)


dcicutils/beanstalk_utils.py
```diff
@@ -668,9 +668,12 @@ def create_bs(envname, load_prod, db_endpoint, es_url, for_indexing=False):
     Returns:
         dict: boto3 res from create_environment/update_environment
     """
+    if is_stg_or_prd_env(envname):
+        raise RuntimeError("beanstalk_utils.create_bs is not approved for production use.")
```
willronchetti, line 672: These functions should really be totally deprecated, I don't think they work at all anymore.

dcicutils/command_utils.py
```diff
@@ -0,0 +1,38 @@
+from .misc_utils import PRINT
+
+
+def yes_or_no(question, quick=None, default=None):
+    """
+    Loops asking a question interactively until it gets a 'yes' or 'no' response. Returns True or False accordingly.
+
+    :param question: The question to ask (without prompts for possible responses, which will be added automatically).
+    :param quick: Whether to allow short-form responses.
```
willronchetti, line 9: I don't love how this interacts with `default`, seems confusing to me. Shouldn't both quick and default be booleans?

test/test_data_utils.py
```diff
@@ -109,3 +111,74 @@ def test_generate_sample_fastq_gzip_content_with_gunzip():
...
+    with pytest.raises(RuntimeError):  # This case is self-contradictory so raises an error.
+        normalize_suffixes("foo.gz", FASTQ_SUFFIXES, compressed=False)
+
+    assert normalize_suffixes("foo.fastq", FASTQ_SUFFIXES) == ("foo.fastq", False)
+    assert normalize_suffixes("foo.fastq", FASTQ_SUFFIXES, compressed=None) == ("foo.fastq", False)
+    assert normalize_suffixes("foo.fastq", FASTQ_SUFFIXES, compressed=False) == ("foo.fastq", False)
+    assert normalize_suffixes("foo.fastq", FASTQ_SUFFIXES, compressed=True) == ("foo.fastq.gz", True)
+
+
+def test_c4_278():
```
willronchetti, line 136: Can we use a better descriptor (or put some info in the docstring)? I think referencing the ticket makes sense though.

dcicutils/command_utils.py
netsettler: I played with it interactively a bunch and it feel stupid to say
```
Do you want to do some horrible thing? (yes/no)
Spell out 'yes' for yes or 'no' for no, or just hit Enter to take the default without spelling anything out.
```

dcicutils/beanstalk_utils.py
netsettler: I'm not sure what "these functions" spans, but I'll write a ticket against that for work another day.

dcicutils/command_utils.py
netsettler: I have a theory of how to fix this.

test/test_data_utils.py
netsettler: Well, I think if we had a convention it might be easier. If the name was `test_jira_c4_278`, would that be enough?

dcicutils/command_utils.py
netsettler: OK, I've separated this into two functions, `yes_or_no` (which takes no arguments beyond the query string) and `y_or_n` (which takes only a default beyond the query string). The original function is renamed to internal-only, so the full generality of it is available but not promoted.

It still has a three-value default for `y_or_n` in order not to presuppose a True/False preference by default.
`>>> y_or_n("foo?")`
`foo? [y/n]:` ⏎
`Please answer 'y' or 'n'.`
`foo? [y/n]:` y⏎
`True`
`>>> y_or_n("foo?", default=True)`
`foo? [Y/n]:` ⏎
`True`
`>>> y_or_n("foo?", default=False)`
`foo? [y/N]:` ⏎
`False`
`>>> y_or_n("foo?", default=None)`
`foo? [y/n]:` ⏎
`Please answer 'y' or 'n'.`
`foo? [y/n]:`

dcicutils/beanstalk_utils.py
netsettler: I made a ticket and put a comment about it in the source.

netsettler:
@willronchetti, since you're out this afternoon, I'm going to go ahead with these changes as I think they're probably close enough to being the kinds of things you intended. Please take a look at this later when you're back and I'll make adjustments as a followup if needed, though I don't anticipate that's a high probability thing.

PR #112, opened 2020-09-14
Miscellaneous utilities ported from cgap-portal and SubmitCGAP repos

netsettler:
This contains functionality that was in the `cgap-portal` and `SubmitCGAP` repo but that we had flagged as better to live in `dcicutils` so it can be shared for use in other places. In a few cases, I renamed or enhanced the feature slightly to be of more general utility. There might be some missing tests since the versions that existed in the other repos were sometimes in test files and test files don't have their own test files. Overall, the testing situation should be improved, though, from where the code was before. It's still fine in review to flag issues. There may be some missing doc strings in some cases, too. But I wanted to get this PR up for review in case there was other stuff I should be attending to.

This is background / low priority. Nothing depends on this happening on any particular schedule. I worked on it this weekend while watching movies. :)

### DIFF
--- a/dcicutils/env_utils.py
+++ b/dcicutils/env_utils.py
@@ -231,6 +231,27 @@ def public_url_mappings(envname):
     return CGAP_PUBLIC_URLS if is_cgap_env(envname) else FF_PUBLIC_URLS
 
 
+def is_cgap_server(server, allow_localhost=False):
+    """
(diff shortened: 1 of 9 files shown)


willronchetti (approved):
I've already looked at most of this so I think it's all good, points about testing are valid though and I trust will be addressed to some extent prior to merging.

PR #113, opened 2020-09-15
Deprecations, updates + CNAME swap

willronchetti:
- Implements an obsolete decorator, applied to many functions in beanstalk_utils
- Fixes some functions in beanstalk_utils that do not work with ES6
- Pull full CNAME swap code from Torb into dcicutils
- TODO: testing?

### DIFF
--- a/dcicutils/beanstalk_utils.py
+++ b/dcicutils/beanstalk_utils.py
@@ -44,14 +44,14 @@ FOURSIGHT_URL = 'https://foursight.4dnucleome.org/'
 FF_MAGIC_CNAME = 'fourfront-green.us-east-1.elasticbeanstalk.com'
 # CGAP_MAGIC_CNAME corresponds to cgap.hms.harvard.edu
 CGAP_MAGIC_CNAME = 'fourfront-cgap.9wzadzju3p.us-east-1.elasticbeanstalk.com'
-# The legacy name MAGIC_CNAME is obsolete (retained for backward compatibility until a major release boundary).
+# The legacy name MAGIC_CNAME is deprecated (retained for backward compatibility until a major release boundary).
 MAGIC_CNAME = FF_MAGIC_CNAME
 
 # FF_GOLDEN_DB is the database behind data.4dnucleome.org (and shared by staging.4dnucleome.org)
 FF_GOLDEN_DB = 'fourfront-production.co3gwj7b7tpq.us-east-1.rds.amazonaws.com'
 # CGAP_GOLDEN_DB is the database behind cgap.hms.harvard.edu
 CGAP_GOLDEN_DB = 'fourfront-cgap.co3gwj7b7tpq.us-east-1.rds.amazonaws.com'
-# The name GOLDEN_DB is obsolete (retained for backward compatibility until a major release boundary).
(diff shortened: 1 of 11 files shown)


dcicutils/beanstalk_utils.py
```diff
@@ -13,7 +13,7 @@
 from datetime import datetime
 from . import ff_utils
 from botocore.exceptions import ClientError
-from .misc_utils import PRINT
+from .misc_utils import PRINT, deprecated
```
netsettler, line 16: I'd like to see this deprecation stuff renamed. Deprecation is advisory and doesn't change semantics. You are making these obsolete, which I endorse.

dcicutils/beanstalk_utils.py
```diff
@@ -223,6 +225,35 @@ def swap_cname(src, dest):
...
+    # delete initial checks (? not clear why this was happening before)
+    if fs['foursight'].get('initial_checks'):
+        del fs['foursight']['initial_checks']
+
+    return fs
+
+
+def swap_cname(src, dest):
+    """ Does a CNAME swap and foursight configuration (pulled in from Torb)
+        NOTE: this is the mechanism by which CNAME swaps must occur now!
```
netsettler, line 246: Now tends to drift. I recommend writing "As of ...date..., this is the preferred mechanism."

dcicutils/beanstalk_utils.py
```diff
@@ -865,7 +904,11 @@ def create_foursight(dest_env, bs_url, es_url, fs_url=None):
     if not bs_url.endswith("/"):
         bs_url += "/"
 
-    es_url = es_url.rstrip(":80")
+    if ':80' in es_url:
+        es_url = es_url.rstrip(":80")
+    elif ':443' in es_url:
+        es_url = es_url.rstrip(":443")
```
netsettler, line 910: You want to be using misc_utils.remove_suffix.  `.rstrip` takes an argument of a set of characters to strip, not a final substring. `.rstrip(':443')` is the same as `.rstrip('43:')` and is not what you want to do.

netsettler (changes requested):
I think we shouldn't fight Python deprecation. I pretty strongly recommend some name changes here (deprecated to obsolete, etc.)

dcicutils/beanstalk_utils.py
```diff
@@ -1252,24 +1254,24 @@ def auth0_client_update(url):
...
-               "client_secret": secret,
-               "audience": "https://hms-dbmi.auth0.com/api/v2/"}
-    headers = {'content-type': "application/json"}
-    res = requests.post("https://hms-dbmi.auth0.com/oauth/token",
+    client = os.environ['Auth0Client']
+    secret = os.environ['Auth0Secret']
+    payload = {'grant_type': 'client_credentials',
+               'client_id': client,
+               'client_secret': secret,
+               'audience': 'https://hms-dbmi.auth0.com/api/v2/'}
```
netsettler, line 1262: Personally, I like `'...'` for tokens and `"..." for free format strings, so I would make URLs in doublequotes. But there's no set rule, so this is fine.

netsettler (approved):
Yeah, this is fine.

PR #114, opened 2020-09-23
Port Some Utility

willronchetti:
- Pulls up some new utility used downstream in foursight/snovault etc into this repository so they can be used elsewhere but imported centrally from here
- Namely, helper functions for common pages/info we'd like to know + `CachedField`

### DIFF
--- a/dcicutils/ff_utils.py
+++ b/dcicutils/ff_utils.py
@@ -1041,7 +1041,7 @@ def get_indexing_status(key=None, ff_env=None):
     return _get_page(page='/indexing_status', key=key, ff_env=ff_env)
 
 
-def are_counts_even(env):
+def get_counts_summary(env):
     """ Returns a named tuple given an FF name to check representing the counts state.
             CountSummary
                 are_even: boolean on whether or not counts are even
--- a/dcicutils/ff_utils.py
+++ b/dcicutils/ff_utils.py
@@ -1041,13 +1041,16 @@ def get_indexing_status(key=None, ff_env=None):
     return _get_page(page='/indexing_status', key=key, ff_env=ff_env)
 
 
+# namedtuple definition used below and can be imported elsewhere
+CountSummary = namedtuple('CountSummary', ['are_even', 'summary_total'])
+
+
 def get_counts_summary(env):
     """ Returns a named tuple given an FF name to check representing the counts state.
             CountSummary
                 are_even: boolean on whether or not counts are even
                 summary_total: raw value of counts
     """
-    CountSummary = namedtuple('CountSummary', ['are_even', 'summary_total'])
     totals = get_counts_page(ff_env=env)
     if 'error' in totals:  # error encountered getting page, assume false and return error
         return CountSummary(are_even=False, summary_total=totals)
--- a/dcicutils/ff_utils.py
+++ b/dcicutils/ff_utils.py
@@ -1010,19 +1010,49 @@ def expand_es_metadata(uuid_list, key=None, ff_env=None, store_frame='raw', add_
     return store, list(item_uuids)
 
 
-def get_health_page(key=None, ff_env=None):
-    """
-    Simple function to return the json for a FF health page given keys or
-    ff_env. Will return json containing an error rather than raising an
-    exception if this fails, since this function should tolerate failure
-    """
+def _get_page(page='/health', key=None, ff_env=None):
+    """ Wrapper for commonly used code to GET a page from an environment
+        Given keys or ff_env, will return json containing an error rather than raising an
+        exception if this fails, since this function should tolerate failure """
     try:
         auth = get_authentication_with_server(key, ff_env)
-        health_res = authorized_request(auth['server'] + '/health', auth=auth, verb='GET')
+        health_res = authorized_request(auth['server'] + page, auth=auth, verb='GET')
         ret = get_response_json(health_res)
     except Exception as exc:
         ret = {'error': str(exc)}
     return ret
+
+
+def get_health_page(key=None, ff_env=None):
+    """
+    Simple function to return the json for a FF health page
+    """
+    return _get_page(page='/health', key=key, ff_env=ff_env)
+
+
+def get_counts_page(key=None, ff_env=None):
+    """ Gets DB/ES counts page in JSON """
+    return _get_page(page='/counts', key=key, ff_env=ff_env)
+
+
+def get_indexing_status(key=None, ff_env=None):
+    """ Gets indexing status counts page in JSON """
+    return _get_page(page='/indexing_status', key=key, ff_env=ff_env)
+
+
+def are_counts_even(env):
+    """ Returns 2-tuple of boolean on whether or not counts are even the value of the counts """
+    totals = get_counts_page(ff_env=env)
+    if 'error' in totals:  # error encountered getting page, assume false and return error
+        return False, totals
+    totals = totals['db_es_total'].split()
+
+    # example value of split totals: ["DB:", "74048", "ES:", "74048"]
+    db_total = totals[1]
+    es_total = totals[3]
+    if int(db_total) > int(es_total):
+        return False, totals
+    return True, totals
 
 
 class SearchESMetadataHandler(object):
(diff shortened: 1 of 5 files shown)


dcicutils/ff_utils.py
```diff
@@ -1010,21 +1010,51 @@ def remove_keys(my_dict, remove_list):
     return store, list(item_uuids)
 
 
-def get_health_page(key=None, ff_env=None):
-    """
-    Simple function to return the json for a FF health page given keys or
-    ff_env. Will return json containing an error rather than raising an
-    exception if this fails, since this function should tolerate failure
-    """
+def _get_page(page='/health', key=None, ff_env=None):
```
netsettler, line 1013: This is a pretty darned weird default. Is there some reason not to just require this argument? I think somehow saying `._get_page()` to mean "get the health page" is just odd.

If you're trying to encourage (well, require) the argument to be a keyword argument, you can do:
```
def _get_page(*, page, key=None, ff_env=None):
```

dcicutils/ff_utils.py
```diff
@@ -1010,21 +1010,51 @@ def remove_keys(my_dict, remove_list):
...
+    """ Wrapper for commonly used code to GET a page from an environment
+        Given keys or ff_env, will return json containing an error rather than raising an
+        exception if this fails, since this function should tolerate failure """
     try:
         auth = get_authentication_with_server(key, ff_env)
-        health_res = authorized_request(auth['server'] + '/health', auth=auth, verb='GET')
+        health_res = authorized_request(auth['server'] + page, auth=auth, verb='GET')
```
netsettler, line 1019: You're making an assumption about the auth not ending in a slash. Would some sort of join function be better? `os.path.join` isn't quite right, but we might have a utility around to join URL parts, or should make one. (Probably the Python url operations have such a thing and I just don't know it.)

dcicutils/ff_utils.py
```diff
@@ -1010,21 +1010,51 @@ def remove_keys(my_dict, remove_list):
...
+    """ Gets DB/ES counts page in JSON """
+    return _get_page(page='/counts', key=key, ff_env=ff_env)
+
+
+def get_indexing_status(key=None, ff_env=None):
+    """ Gets indexing status counts page in JSON """
+    return _get_page(page='/indexing_status', key=key, ff_env=ff_env)
+
+
+def are_counts_even(env):
```
netsettler, line 1043: This will always return a truthy value so should not have a name that looks like a predicate. Maybe call it `check_counts_even` or `counts_summary` or something like that.

netsettler: This might be a place where returning a `namedtuple` would be a good idea so the person receiving it can access the parts by name instead of by position if they want only one, e.g., if the parts were `are_even` and `summary_total` then `check_counts(...).are_even` serves as a predicate.

dcicutils/ff_utils.py
```diff
@@ -1010,21 +1010,51 @@ def remove_keys(my_dict, remove_list):
...
+    if 'error' in totals:  # error encountered getting page, assume false and return error
+        return False, totals
+    totals = totals['db_es_total'].split()
+
+    # example value of split totals: ["DB:", "74048", "ES:", "74048"]
+    db_total = totals[1]
+    es_total = totals[3]
+    if int(db_total) > int(es_total):
+        return False, totals
+    return True, totals
```
netsettler, line 1055: Do you really not want to just return `True, db_total, es_total` here? Is the caller of this function likely to want to use the labels?

test/test_ff_utils.py
```diff
@@ -285,7 +285,7 @@ def test_stuff_in_queues(integrated_ff):
     # just take the first handful
     for item in search_res[:8]:
         ff_utils.patch_metadata({}, obj_id=item['uuid'], key=integrated_ff['ff_key'])
-    time.sleep(5)  # let queues catch up
+    time.sleep(3)  # let queues catch up
```
netsettler, line 288: I wish we did something more abstract like `let_queues_catch_up()` even if it was just implemented as a sleep, though I suppose we do different amounts of sleep but I suppose that's a problem for another day. What you're doing here is pretty common.

test/test_ff_utils.py
```diff
@@ -1148,3 +1149,25 @@ def test_convert_param():
...
+    assert 'ES' in counts_res
+    indexing_status_res = ff_utils.get_indexing_status(ff_env=ff_env)
+    assert 'primary_waiting' in indexing_status_res
+
+
+@pytest.mark.integrated
+def test_are_counts_even(integrated_ff):
+    ff_env = integrated_ff['ff_env']
+    counts_are_even, totals = ff_utils.are_counts_even(ff_env)
+    if counts_are_even:
```
netsettler, line 1170: In that function where it says
```
# example value of split totals: ["DB:", "74048", "ES:", "74048"]
```
it woul be helpful to show an example where 'more items' shows up.

test/test_misc_utils.py
```diff
@@ -1156,3 +1156,115 @@ def test_file_contents():
         assert file_contents("foo.bin", binary=True) == b'\x48\x65\x6c\x6c\x6f\x21\x0a'
         assert file_contents("foo.bin", binary=False) == b'\x48\x65\x6c\x6c\x6f\x21\x0a'.decode('utf-8')
         assert file_contents("foo.bin", binary=False) == 'Hello!\n'
+
+
+def make_counter(start=0):
```
netsettler, line 1161: Now that we're in dcicutils, we might want to move this function to `misc_utils` or `qa_utils` so it's available to other things.

test/test_misc_utils.py
```diff
@@ -1156,3 +1156,115 @@ def test_file_contents():
...
+
+    def test_cached_field_timeout(self):
+        field = CachedField('simple1', update_function=make_counter())
+
+        assert field.timeout == self.DEFAULT_TIMEOUT
+        field.set_timeout(30)
+        assert field.timeout == 30
+
+
+@pytest.mark.parametrize('input, expected', [
```
netsettler, line 1261: Probably it should have something like `XMLContainer` and `X_M_L_Container` to emphasize the weird thing that is likely to happen there even if it follows from the ABCD example. Mostly.

netsettler (approved):
This looks generally reasonable, even though I have a number of smallish things I wanted to call out.

dcicutils/ff_utils.py
```diff
@@ -1040,19 +1041,25 @@ def get_indexing_status(key=None, ff_env=None):
...
 
-def are_counts_even(env):
-    """ Returns 2-tuple of boolean on whether or not counts are even the value of the counts """
+def get_counts_summary(env):
+    """ Returns a named tuple given an FF name to check representing the counts state.
+            CountSummary
+                are_even: boolean on whether or not counts are even
+                summary_total: raw value of counts
+    """
+    CountSummary = namedtuple('CountSummary', ['are_even', 'summary_total'])
```
netsettler, line 1050: I would probably put this at toplevel. It is effectively a type declaration. If you do it this way, you'll be going through the whole labor of defining and using a disposable type on every call (which actually does string manipulation and calls eval or execute to define the class, so can't easily be optimized by the JIT unless it has special knowledge of this class). But basically it means the type of any return value will not match the type of any other return value, so it's a bit of a hassle for the GC, too. Sorry if my earlier example confused you on this.

dcicutils/misc_utils.py
```diff
@@ -802,3 +812,14 @@ def __repr__(self):
         return 'CachedField %s with update function %s on timeout %s' % (
             self.name, self._update_function, self.timeout
         )
+
+
+def make_counter(start=0):
```
netsettler, line 817: Might be worth adding `, step=1` here in the params, and later saying `storage[0] += step`. Would require more test cases. We could add this later, but it's such an obvious detail this might be the better time.

For number theory fans, it's worth noting that if you add a `step`, you can also do odd things like
```
make_counter([], step=['foo'])
```
or
```
make_counter("", step=". ")
```
not that I imagine anyone will.

netsettler (approved):
I don't see any blockers here.

PR #115, opened 2020-10-08
Miscellaneous fixes 2020-10-06

netsettler:
You _could_ just see the CHANGELOG.md file for info. 🙂  (I think this can be a .rst file without any problem, too, and will probably change it. I was being lazy.) But here we go:

 * Fix a lurking bug in `beanstalk_utils` where `delete_db` had the wrong scope.
 * Add `qa_utils.raises_regexp` for conceptual compatibility with `AssertRaises` in `unittest`.
 * Add `qa_utils.CustomizableProperty` and companion `getattr_customized`.
 * Add `qa_utils.override_dict`, factored out of `qa_utils.override_environ`.
 * Add `qa_utils.check_duplicated_items_by_key` to aid in error reporting for search results.
 * Add support for `sentry_dsn` and a `ENCODED_SENTRY_DSN`  beanstalk environment variable in `deployment_utils`.
 * Add `qa_utils.MockUUIDModule` for being able to mock `uuid.uuid4()`.

I also fixed the buggy test case for `ff_utils.search_metadata` so that tests would pass. This is not just about an intermittent but actually suggests this function has marginal behavior in the presence of such transients. It might be we could design it better.  I believe the problem is that it's pulling in results in pieces (chunk size 50, as it happened) and there are inconsistencies between those grabs for data. If it were expecting 9 items ABCDEFGHI and was taking them 3 at a time, it's behaving as if because indexing is ongoing, only 8 items are available at the time it runs the first chunk, but those items are ACDEFGHI, so it gets ACD on the first try and then later when B appears and it calls it a second time it gets DEF instead of BEF, so it misses B and adds an extra D, and the ultimate result is ABDDEFGHI and has two D's and no B. I can show via the unit test technology that this is so (though doing so would fail the tests). We need to decide if we want this fixed _and how_, and then I can figure out what to do.


test/test_qa_utils.py
```diff
@@ -1048,3 +1115,256 @@ def test_mock_keys_not_implemented():
...
+    FAVORITE_SONG = 'Jingle Bells'
+
+
+def test_find_field_declared_class():
+
+    thing = SampleClass2(favorite_fruit='orange')
+
+    [kind, value] = UncustomizedInstance._find_field_declared_class(thing, 'favorite_fruit')
+    assert kind == 'instance'
+    assert value == 'orange'
```
willronchetti, line 1145: So use case here is to see which class in a potentially large number of parent classes does a property belong to?

test/test_qa_utils.py
```diff
@@ -1048,3 +1115,256 @@ def test_mock_keys_not_implemented():
...
+
+def test_uncustomized_instance():
+
+    thing = SampleClass2(favorite_fruit=CustomizableProperty('favorite_fruit',
+                                                             description="the string name of a fruit"))
+
+    assert str(UncustomizedInstance(thing, field='FAVORITE_SONG')) == (
+        "Attempt to access field FAVORITE_SONG from class test.test_qa_utils.SampleClass2."
+        " It was expected to be given a custom value in a subclass: the string name of a song."
+    )
```
willronchetti, line 1169: Is `UncustomizedInstance` intended for use in debugging? Can you maybe describe some scenarios where you'd want to use this? I don't feel like I'm getting it from the tests here.

test/test_qa_utils.py
```diff
@@ -1048,3 +1115,256 @@ def test_mock_keys_not_implemented():
...
+
+    with mock.patch("os.path.exists", mocked_exists):
+
+        class MyVersionChecker(VersionChecker):
+
+            PYPROJECT = os.path.join(os.path.dirname(__file__), "../pyproject.toml")
+            CHANGELOG = os.path.join(os.path.dirname(__file__), "../CHANGELOG.md")
+
+        with pytest.raises(AssertionError):
+            MyVersionChecker.check_version()  # The version history will be missing because of mocking.
```
willronchetti, line 1356: I'm mixed on whether I think things like this should throw assertion error, despite it being used in QA.

willronchetti (approved):
I wouldn't say I have a problem with any of this, but I certainly feel like I don't totally understand the use case for some of the new `qa_utils` functionality. Some explanation/documentation there I think would help. Other than that LGTM.

test/test_ff_utils.py
```diff
@@ -471,28 +479,77 @@ def test_upsert_metadata(integrated_ff):
...
     assert len(search_res_slash) == len(search_res)
     # search with a limit
     search_res_limit = ff_utils.search_metadata(url + '/search/?limit=3&type=File', key=integrated_ff['ff_key'])
     assert len(search_res_limit) == 3
     # search with a limit from a certain entry
-    search_res_from_limit = ff_utils.search_metadata(url + '/search/?type=File&from=5&limit=53', key=integrated_ff['ff_key'])
+    search_res_from_limit = ff_utils.search_metadata(url + '/search/?type=File&from=5&limit=53',
+                                                     key=integrated_ff['ff_key'])
     assert len(search_res_from_limit) == 53
     # search with a filter
```
aschroed, line 554: Is this actually doing any filtering since it's using the mock result?

aschroed (approved):
I see what you did with the mocking of search result and what the issue is that motivated the unit test.  It would indeed be nice to be able to do something sensible so search results could somehow provide a warning if the issue that prompted this came up in an actual search or alternatively to somehow prevent that from happening at all.

dcicutils/qa_utils.py
```diff
@@ -551,8 +559,150 @@ def download_fileobj(self, Bucket, Key, Fileobj, **kwargs):  # noqa - Uppercase
...
+        # That's not an intended use case, but just report it without involving mention of the class.
+        raise UncustomizedInstance(instance=thing, field=key)
+    else:
+        return value
+
+
+class VersionChecker:
+
+    PYPROJECT = CustomizableProperty('PYPROJECT', description="The repository-relative name of the pyproject file.")
+    CHANGELOG = CustomizableProperty('CHANGELOG', description="The repository-relative name of the change log.")
```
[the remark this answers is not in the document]
netsettler: It supports the ability to publish abstract classes in one library and to get decent error messages when they're used wrong in another library. I find Python does this poorly.  I might like more of our repos to use the version checker, for example, in order to promote use of changelogs. But I'm thinking it could be useful for other things that snovault might want to publish to FF and CGAP, too, so it seemed like it might be worth bundling in. It's a bit speculative to do on basis of a single use here, though I found in prior work elsewhere that once I had this facility, I used it a lot and it gave better error messages than Python presently gives if you just either use an attribute that requires complex subclass initialization across libraries or you inline the error checks every time you do. I could leave it out until we have more need if you thought that advisable. It's used in the VersionChecker class if you want to see a use.

dcicutils/qa_utils.py
```diff
@@ -551,8 +559,150 @@ def download_fileobj(self, Bucket, Key, Fileobj, **kwargs):  # noqa - Uppercase
...
+def raises_regexp(error_class, pattern):
+    # Mostly compatible with unittest style, so that (approximately):
+    #  pytest.raises(error_class, match=exp) == unittest.TestCase.assertRaisesRegexp(error_class, exp)
+    # They differ on what to do if an error_class other than the expected one is raised.
+    return pytest.raises(error_class, match=pattern)
+
+
+def check_duplicated_items_by_key(key, items, url=None, formatter=str):
+
+    __tracebackhide__ = True
```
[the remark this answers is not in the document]
netsettler: In PyTest it causes the the function with this in it not to be the focus of the error message. Instead its caller is treated as the one that had the error, as if this function were internal to 'raise'. Our `check_true` function should have this, for example. I just ran across it when trying to write `raises_regexp` but then ended up solving this a different way because it turned out `pytest` had the functionality built in (and implemented using this feature).

test/test_ff_utils.py
```diff
@@ -471,28 +479,77 @@ def test_upsert_metadata(integrated_ff):
...
+def test_search_metadata_unit(integrated_ff, url):
+    with mock.patch.object(ff_utils, "authorized_request", make_mocked_search()):
+        check_search_metadata(integrated_ff, url)
+
+
+# @pytest.mark.integrated
+# @pytest.mark.flaky
+# @pytest.mark.parametrize('url', ['', 'to_become_full_url'])
+# def test_search_metadata_integrated(integrated_ff, url):
+#     check_search_metadata(integrated_ff, url)
```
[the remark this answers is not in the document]
netsettler: I'll discuss this with you offline and then act on it as needed here, but thanks for flagging it so I can remember to do that.

test/test_qa_utils.py
netsettler: Yeah, it tells you whose documentation you need to read.

test/test_qa_utils.py
netsettler: It's intended to help you get a decent error message at the right time when you reasonably know that you're only half-implementing something (which is what an abstract class is - a half-implementation). Otherwise you get inscrutable error messages that cost debugging time.

test/test_qa_utils.py
netsettler: What would you use instead?

test/test_ff_utils.py
netsettler: It is not doing type filtering. It is giving back a different number of items based on what you do. The thing that's mocked is the authorized_request call. So if you run this test with suitable verbosity, you'll see this output (where the "..." and the "#" comments are not me abbreviating here, but actually the result of the output since I thought otherwise the output was overly verbose and impeded my debugging):
```
mocked search http://fourfront-mastertest.9wzadzju3p.us-east-1.elasticbeanstalk.com/search/?limit=50&type=File&sort=-date_created&from=0
yields Search [{'uuid': '0'}, ..., {'uuid': '49'}] # 50 items
mocked search http://fourfront-mastertest.9wzadzju3p.us-east-1.elasticbeanstalk.com/search/?limit=50&type=File&sort=-date_created&from=50
yields Search [{'uuid': '50'}, ..., {'uuid': '99'}] # 50 items
mocked search http://fourfront-mastertest.9wzadzju3p.us-east-1.elasticbeanstalk.com/search/?limit=50&type=File&sort=-date_created&from=100
yields Search [{'uuid': '100'}, ..., {'uuid': '129'}] # 30 items
mocked search http://fourfront-mastertest.9wzadzju3p.us-east-1.elasticbeanstalk.com/search/?limit=50&type=File&sort=-date_created&from=0
yields Search [{'uuid': '0'}, ..., {'uuid': '49'}] # 50 items
mocked search http://fourfront-mastertest.9wzadzju3p.us-east-1.elasticbeanstalk.com/search/?limit=50&type=File&sort=-date_created&from=50
yields Search [{'uuid': '50'}, ..., {'uuid': '99'}] # 50 items
mocked search http://fourfront-mastertest.9wzadzju3p.us-east-1.elasticbeanstalk.com/search/?limit=50&type=File&sort=-date_created&from=100
yields Search [{'uuid': '100'}, ..., {'uuid': '129'}] # 30 items
mocked search http://fourfront-mastertest.9wzadzju3p.us-east-1.elasticbeanstalk.com/search/?limit=50&type=File&sort=-date_created&from=0
yields Search [{'uuid': '0'}, ..., {'uuid': '49'}] # 50 items
mocked search http://fourfront-mastertest.9wzadzju3p.us-east-1.elasticbeanstalk.com/search/?type=File&from=5&limit=50&sort=-date_created
yields Search [{'uuid': '5'}, ..., {'uuid': '54'}] # 50 items
mocked search http://fourfront-mastertest.9wzadzju3p.us-east-1.elasticbeanstalk.com/search/?type=File&from=55&limit=50&sort=-date_created
yields Search [{'uuid': '55'}, ..., {'uuid': '104'}] # 50 items
mocked search http://fourfront-mastertest.9wzadzju3p.us-east-1.elasticbeanstalk.com/search/?limit=50&type=File&file_type=reads&sort=-date_created&from=0
yields Search [{'uuid': '0'}, ..., {'uuid': '49'}] # 50 items
mocked search http://fourfront-mastertest.9wzadzju3p.us-east-1.elasticbeanstalk.com/search/?limit=50&type=File&file_type=reads&sort=-date_created&from=0
yields Search [{'uuid': '0'}, ..., {'uuid': '49'}] # 50 items
mocked search http://fourfront-mastertest.9wzadzju3p.us-east-1.elasticbeanstalk.com/browse/?limit=50&type=File&sort=-date_created&from=0
yields Browse [{'uuid': '0'}, ..., {'uuid': '49'}] # 50 items
PASSED   [100%]mocked search http://fourfront-mastertest.9wzadzju3p.us-east-1.elasticbeanstalk.com/search/?limit=50&type=File&sort=-date_created&from=0
yields Search [{'uuid': '0'}, ..., {'uuid': '49'}] # 50 items
mocked search http://fourfront-mastertest.9wzadzju3p.us-east-1.elasticbeanstalk.com/search/?limit=50&type=File&sort=-date_created&from=50
yields Search [{'uuid': '50'}, ..., {'uuid': '99'}] # 50 items
mocked search http://fourfront-mastertest.9wzadzju3p.us-east-1.elasticbeanstalk.com/search/?limit=50&type=File&sort=-date_created&from=100
yields Search [{'uuid': '100'}, ..., {'uuid': '129'}] # 30 items
mocked search http://fourfront-mastertest.9wzadzju3p.us-east-1.elasticbeanstalk.com//search/?limit=50&type=File&sort=-date_created&from=0
yields Search [{'uuid': '0'}, ..., {'uuid': '49'}] # 50 items
mocked search http://fourfront-mastertest.9wzadzju3p.us-east-1.elasticbeanstalk.com//search/?limit=50&type=File&sort=-date_created&from=50
yields Search [{'uuid': '50'}, ..., {'uuid': '99'}] # 50 items
mocked search http://fourfront-mastertest.9wzadzju3p.us-east-1.elasticbeanstalk.com//search/?limit=50&type=File&sort=-date_created&from=100
yields Search [{'uuid': '100'}, ..., {'uuid': '129'}] # 30 items
mocked search http://fourfront-mastertest.9wzadzju3p.us-east-1.elasticbeanstalk.com//search/?limit=50&type=File&sort=-date_created&from=0
yields Search [{'uuid': '0'}, ..., {'uuid': '49'}] # 50 items
mocked search http://fourfront-mastertest.9wzadzju3p.us-east-1.elasticbeanstalk.com//search/?type=File&from=5&limit=50&sort=-date_created
yields Search [{'uuid': '5'}, ..., {'uuid': '54'}] # 50 items
mocked search http://fourfront-mastertest.9wzadzju3p.us-east-1.elasticbeanstalk.com//search/?type=File&from=55&limit=50&sort=-date_created
yields Search [{'uuid': '55'}, ..., {'uuid': '104'}] # 50 items
mocked search http://fourfront-mastertest.9wzadzju3p.us-east-1.elasticbeanstalk.com//search/?limit=50&type=File&file_type=reads&sort=-date_created&from=0
yields Search [{'uuid': '0'}, ..., {'uuid': '49'}] # 50 items
mocked search http://fourfront-mastertest.9wzadzju3p.us-east-1.elasticbeanstalk.com//search/?limit=50&type=File&file_type=reads&sort=-date_created&from=0
yields Search [{'uuid': '0'}, ..., {'uuid': '49'}] # 50 items
mocked search http://fourfront-mastertest.9wzadzju3p.us-east-1.elasticbeanstalk.com//browse/?limit=50&type=File&sort=-date_created&from=0
yields Browse [{'uuid': '0'}, ..., {'uuid': '49'}] # 50 items
```
I didn't see the test testing for any additional kind of filtering. I was trying to convert the test from an integration test to a unit test. The assumption in unit testing is that you only test the new layer of functionality and that the underlying stuff (in this case the search endpoint) is tested already elsewhere. So instead this just checks that the arguments are always what it expects so that if you start asking for different values, it's easier to see that the mock needs to be extended. Does that make sense? Is there some additional functionality that you don't think is tested?

A separate and reasonable question you might be asking is: Should we also have a set of integration tests or system tests that test the search functionality from the outside. The answer to that question might be yes. If you mean this to actually be a test of /search/ itself, that may be important. But that seems a separate thing from testing this endpoint.

netsettler: And there's a more subtle question: Really this functionality seems to stitch together the result of multiple paginated calls. As far as I can tell, if ES (due to indexing) is changing underneath, it won't get that right.  You will sometimes omit results. That problem is unsolvable probably. You can't know they are coming. But sometimes you'll get duplications. In that case, you could just recognize that it's an artifact of something getting introduced. It'd be expensive to back up and redo the calls, but minimally it might be good to make a pass through and de-dup the result rather than returning the same item more than once.

aschroed: I understand what a unit test is and what you are testing and my comment is just pointing out that the additional search with the file_type filter can be removed as it's really not doing anything.

willronchetti (approved):

test/test_ff_utils.py
netsettler: Well, when I reenable the part that's the integration test (which I plan to do one I have the integration and unit tests more fully separated) then it will be needed to keep the test small. So I think it's OK to leave, but I take your point. (Also, it somewhat has the vague purpose of making the point that we can't just look for some constant set of query args because there will be others littered in there.)

PR #116, opened 2020-10-09
Port quick_deepcopy as copy_json

willronchetti:
- Brings in the `quick_deepcopy` function from https://github.com/ENCODE-DCC/snovault/blob/d14ae8cb4e3bf4bb76ade583f914ec7283a91379/src/snovault/util.py#L91 and renames it `copy_json`
- Add some basic tests

### DIFF
--- a/test/test_misc_utils.py
+++ b/test/test_misc_utils.py
@@ -16,6 +16,7 @@ from dcicutils.misc_utils import (
     Retry, apply_dict_overrides, utc_today_str, RateManager, environ_bool,
     LockoutManager, check_true, remove_prefix, remove_suffix, full_class_name, full_object_name, constantly,
     keyword_as_title, file_contents, CachedField, camel_case_to_snake_case, snake_case_to_camel_case, make_counter,
+    copy_json,
 )
 from dcicutils.qa_utils import Occasionally, ControlledTime, override_environ, MockFileSystem
 from unittest import mock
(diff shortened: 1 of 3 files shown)


test/test_misc_utils.py
```diff
@@ -1282,3 +1283,20 @@ def test_camel_case_to_snake_case(input, expected):
...
+    {'list2': [1, 2, 3]},
+    {'list_of_objects': [
+        {'hello': 'world', 'foo': 'bar'},
+        {'hello': 'dog', 'foo': 'cat'}
+    ]},
+    {'object': {'of objects': {'of more objects': {'even more': 'and more'}}}}
+])
+def test_copy_json(obj):
+    """ Tests some basic cases for copy_json """
+    assert copy_json(obj) == obj
```
netsettler, line 1302: I might prefer another test like (untested):
```
def test_copy_json_side_effects():
    obj = {'foo': [1, 2, 3], 'bar': [{'x': 4, 'y': 5}, {'x': 2, 'y': 7}]}
    obj_copy = copy_json(obj)
    obj['foo'][1] = 20
    obj['bar'][0]['y'] = 500
    obj['bar'][1] = 17
    assert obj == {'foo': [1, 20, 3], 'bar': [{'x': 4, 'y': 500}, 17]}
    assert obj_copy == {'foo': [1, 2, 3], 'bar': [{'x': 4, 'y': 5}, {'x': 2, 'y': 7}]}
```

netsettler (approved):

PR #118, opened 2020-11-06
Various bits of functionality in support of 4dn-status (C4-363)

netsettler:
@willronchetti suggested bits of functionality in a recent 4dn-status PR (in the brig repo) should move here. This is mostly the fulfillment of that suggestion. I made some better names and added some testing.

* In `qa_utils`, this makes a `ControlledTime` instance be someone one can use as a mock for the `datetime` module itself, by giving it a `datetime` property that is sort of like the `datetime.datetime` class, though obviously it is not the _actual_ class so mocking has to be done with care. Some code wanting to use such mocks may want to receive that type via alternate means so that the function `datetime.datetime` can be mocked without affecting the type `datetime.datetime`. Such are mocks.
* In `misc_utils`, this adds:
  * `as_seconds` so that, for example `as_seconds(minutes=3)` can be used to get 180.
  * `hms_now` to get the value of `datetime.datetime.now()` in HMS local time (EST or EDT as appropriate).
  * `in_datetime_interval` to test that a given time is within a given time interval.
  * `as_datetime` to coerce a properly formatted `str` to a `datetime.datetime`.

### DIFF
--- a/dcicutils/misc_utils.py
+++ b/dcicutils/misc_utils.py
@@ -444,9 +444,11 @@ HMS_TZ = pytz.timezone("US/Eastern")
 
 def as_datetime(dt, tz=None):
     """
-    Parses the given datetime, returning a datetime.datetime object.
+    Parses the given date/time (which may be a string or a datetime.datetime), returning a datetime.datetime object.
 
-    If the given datetime is already such an object, it is returned.
+    If the given datetime is already such an object, it is just returned.
+    If it is a string, it should be in a format such as 'yyyy-mm-dd hh:mm:ss' or 'yyyy-mm-dd hh:mm:ss-nnnn'
+    (with -nnnn being a timezone specification).
     """
     if dt is None:
         return None
--- a/dcicutils/misc_utils.py
+++ b/dcicutils/misc_utils.py
@@ -5,13 +5,17 @@ This file contains functions that might be generally useful.
 import contextlib
 import datetime
 import functools
+import math
 import io
 import os
 import logging
+import pytz
 import time
 import warnings
 import webtest  # importing the library makes it easier to mock testing
 
+from dateutil.parser import parse as dateutil_parse
+from datetime import datetime as datetime_type
 
 # Is this the right place for this? I feel like this should be done in an application, not a library.
 # -kmp 27-Apr-2020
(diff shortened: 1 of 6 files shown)


dcicutils/misc_utils.py
```diff
@@ -412,6 +416,68 @@ def utc_today_str():
...
+    try:
+        # This type check has to work even if datetime is mocked, so we use it under another variable name to
+        # make it harder to mock out. -kmp 6-Nov-2020
+        if not isinstance(dt, datetime_type):
+            dt = dateutil_parse(dt)
+        if tz and not dt.tzinfo:
+            dt = tz.localize(dt)
+        return dt
+    except Exception:
+        return None
```
willronchetti, line 462: It's not clear to me from the definition what the utility of this function is... If it's already a datetime, how is it not a datetime.datetime object? It seems like you're doing all this casting from datetime objects that is unnecessary... If `dt` were a string this would make sense to me...

willronchetti (approved):
One small comment (that I think is because I'm missing a detail here) - also I think the misc_utils tests could use a little more documentation but I'll leave that up to you.

dcicutils/misc_utils.py
netsettler: It takes a string as an argument. Accepting a datetime also is just for convenience so that other functions can accept times in either parsed or unparsed form. Look to line 457 where it parse a time string into a datetime.

netsettler: I guess I need to make it clearer in the doc string. To me, anything that uses the verb parses means it takes a string (or I guess sometimes a stream, though not here), so the idea that you would think it only took a datetime didn't enter my mind. :)

netsettler: If you look at the tests, you'll see the intended uses:
```
def test_as_datetime():

    t0 = datetime_module.datetime(2015, 7, 4, 12, 0, 0)
    t0_utc = pytz.UTC.localize(datetime_module.datetime(2015, 7, 4, 12, 0, 0))

    assert as_datetime(t0) == t0
    assert as_datetime(t0, tz=pytz.UTC) == t0_utc
    assert not as_datetime('2015-07-04T12:00:00').tzinfo
    assert as_datetime('2015-07-04T12:00:00') == t0
    assert as_datetime('2015-07-04 12:00:00') == t0
    assert as_datetime('2015-07-04 12:00:00', tz=pytz.UTC) == t0_utc
    assert as_datetime('2015-07-04 12:00:00Z').tzinfo
    assert as_datetime('2015-07-04 12:00:00Z') != t0
    assert as_datetime('2015-07-04 12:00:00Z') == t0_utc
    assert as_datetime('2015-07-04 12:00:00-0000') == t0_utc
```

PR #122, opened 2020-11-25
Speed up ff_utils unit tests, and misc small bits of functionality

netsettler:
This makes the unit tests for `dcicutils.ff_utils` run faster, and makes other incidental small changes.

The general approach is to keep existing unit tests of `ff_utils`, most of which are really integration tests, but to add a real unit test that is faster and more comprehensive. Where such unit tests get written, the integration test will be marked `integrationx` instead of `integration`, and such tests will be excluded from normal runs by `make test`, speeding things up. We can arrange another mechanism for periodically running just integration tests or all the tests, or something like that, but I wasn't focused on that for now.

See the `CHANGLOG.rst` for details.  Note that at the time the PR was created, this doesn't make everything a unit test, but it's work in progress.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,10 +6,56 @@ dcicutils
 Change Log
 ----------
 
+1.6.0
+=====
+
+* Added an ``integratedx`` mark to possible marks in ``pytest.ini``. These
+  are the same as ``integrated`` but they represent test cases that have
+  an associated unit test that is redundant, so that the ``integratedx``
+  test doesn't have to be run to get full coverage.
+
+* For ``ff_utils``:
+
+  * Split tests into a ``xxx_unit`` and
+    ``xxx_integrated`` version.  The latter is marked with new
+    ``integratedx`` mark.
+
+* For ``env_utils``:
+
+  * Added some test cases.
+
+* For ``s3_utils``:
+
+  * Small remodularization of ``s3Utils`` for easier access to
+    some constants in testing.
+  * Improvements to error reporting in ``s3Utils.get_access_keys()``.
+
+* For ``qa_utils``:
+
+  * In ``MockFileSystem``, fixed a typo in debugging typeout.
+  * In ``MockResponse``:
+
+    * Added a ``url=`` init arg and ``.url`` property.
+    * Added a .text as synonym for ``.content``.
+
+  * In ``MockBotoS3Client``:
+
+    * Extended to handle ``region_name=``.
+    * Added ``mock_other_required_arguments=`` and ``mock_s3_files=``
+      init args for use in testing.
+    * Added ``MockBotoS3Client``, add ``.get_object(Bucket, Key)``.
+
+* For ``ff_utils``:
+
+  * Used ``ValueError`` rather than ``Exception`` in several
+    places errors are raised.
+  * Some very small other refactoring was also done
+    for modularity that should not affect behavior.
+
 1.5.1
 =====
 
-**PR 120: Update ES-py Version
+**PR 120: Update ES-py Version**
 
 * Updates elasticsearch library to 6.8.1 to take a bug fix.
 
@@ -19,7 +65,7 @@ Change Log
 
 **PR 119: More env_utils support**
 
-* Add ``env_utils.classify_server_url`.
+* Add ``env_utils.classify_server_url``.
 
 
 1.4.0
--- a/dcicutils/ff_utils.py
+++ b/dcicutils/ff_utils.py
@@ -184,6 +184,15 @@ def purge_request_with_retries(request_fxn, url, auth, verb, **kwargs):
     return final_res
 
 
+REQUESTS_VERBS = {
+    'GET': requests.get,
+    'POST': requests.post,
+    'PATCH': requests.patch,
+    'PUT': requests.put,
+    'DELETE': requests.delete,
+}
+
+
 def authorized_request(url, auth=None, ff_env=None, verb='GET',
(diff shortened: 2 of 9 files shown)


CHANGELOG.rst
```diff
@@ -6,10 +6,56 @@ dcicutils
...
+* Added an ``integratedx`` mark to possible marks in ``pytest.ini``. These
+  are the same as ``integrated`` but they represent test cases that have
+  an associated unit test that is redundant, so that the ``integratedx``
+  test doesn't have to be run to get full coverage.
+
+* For ``ff_utils``:
+
+  * Split tests into a ``xxx_unit`` and
+    ``xxx_integrated`` version.  The latter is marked with new
+    ``integratedx`` mark.
```
willronchetti, line 21: In other repos we just use `integrated` - is there any reason for the `x`?

dcicutils/ff_utils.py
```diff
@@ -1145,8 +1154,8 @@ def unified_authentication(auth=None, ff_env=None):
     elif isinstance(auth, tuple) and len(auth) == 2:
         use_auth = auth
     if not use_auth:
-        raise Exception("Must provide a valid authorization key or ff "
-                        "environment. You gave: %s (key), %s (ff_env)" % (auth, ff_env))
+        raise ValueError("Must provide a valid authorization key or ff environment."
```
willronchetti, line 1157: Might consider adding some custom exceptions in the library? Definitely an improvement over `Exception` though.

pytest.ini
```diff
@@ -1,5 +1,6 @@
 [pytest]
 markers =
   working: test should work
-  integrated: this is an integrated test
+  integrated: this is an integration test
+  integratedx: this integration test redundantly tests functionality covered in unit tests
```
willronchetti, line 5: Is this the difference? `integratedx` is redundant?

test/test_ff_utils.py
```diff
@@ -227,12 +235,114 @@ def get_access_keys(self):  # noqa - this is a mock so PyCharm shouldn't suggest
...
+    'server': bar_env_url,
+}
+bar_env_default_auth_dict = {'default': bar_env_auth_dict}
+
+some_server = 'http://localhost:8000'
+some_auth_key, some_auth_secret = some_auth_tuple = ('mykey', 'mysecret')
+some_badly_formed_auth_tuple = ('mykey', 'mysecret', 'other-junk')  # contains an extra element
+some_auth_dict = {'key': some_auth_key, 'secret': some_auth_secret}
+some_auth_dict_with_server = {'key': some_auth_key, 'secret': some_auth_secret, 'server': some_server}
+some_badly_formed_auth_dict = {'kee': some_auth_key, 'secret': some_auth_secret}  # 'key' is misspelled
```
willronchetti, line 264: Seems like these things belong in a fixture or something... They feel out of place here.

test/test_ff_utils.py
```diff
@@ -246,30 +356,68 @@ def test_unified_authentication(integrated_ff):
...
+            # so it will be rejected as bad data by .get_authentication_with_server().
+            ff_utils.get_authentication_with_server(foo_env_auth_dict, None)
+
+        with raises_regexp(ValueError, 'ERROR GETTING SERVER'):
+            # If the auth is not provided (or None), the ff_env must be given (and not None).
+            ff_utils.get_authentication_with_server(None, None)
+
+
+@pytest.mark.integratedx
+@pytest.mark.flaky  # TODO: Why is this marked flaky? There should be nothing flaky about this. -kmp 24-Oct-2020
```
willronchetti, line 419: There are certain operations that can occur that can make this test fail - for example if a deployment or CNAME swap occurs during the run.

test/test_ff_utils.py
```diff
@@ -378,9 +526,90 @@ def test_authorized_request_integrated(integrated_ff):
     assert 'Bad status code' in str(exec_info.value)
 
 
-@pytest.mark.integrated
+def test_get_metadata_unit():
```
willronchetti, line 529: There is a lot going on in this test, a high level description up here may be useful.

willronchetti (approved):
Approving with some small comments

pytest.ini
netsettler: I fixed the doc in pytest.ini to hopefully be more clear:
```
  integrated: an integration test
  integratedx: an excludable integration test, redundantly testing functionality also covered by a unit test
```
I also fixed the make targets in the `Makefile` to hopefully be more clear:
```
test:  # runs default tests, which are the unit tests
	make test-units

test-all:
	pytest -vv

test-units:  # runs unit tests (and integration tests not backed by a unit test)
	pytest -vv -m "not integratedx"

test-integrations:  # runs integration tests
	pytest -vv -m "integrated or integratedx"
```

test/test_ff_utils.py
netsettler: It's almost an anti-pattern here because putting them in a fixture would mean I'd have to unpack them from the fixture. The whole point is just to have these values set up and ready to use by several tests. However, I did some light rearranging and added some comments explaining, so maybe it won't be so jarring.

test/test_ff_utils.py
netsettler: Thanks. I removed the TODO and marked it this way, which I think we should probably do more often:
```
@pytest.mark.flaky  # e.g., if a deployment or CNAME swap occurs during the run
```

test/test_ff_utils.py
netsettler: I rearranged things slightly and added more comments.

aschroed (approved):

PR #123, opened 2020-12-02
C4-402 Add GA for build

willronchetti:
- Configures a new build for use with Github Actions


.github/workflows/main-publish.yml
```diff
@@ -0,0 +1,32 @@
...
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - uses: actions/checkout@v2
+      - name: Publish
+        env:
+          PYPI_USER: ${{ secrets.PYPI_USER }}
+          PYPI_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
+        run: poetry publish --build --username=$PYPI_USER --password=$PYPI_PASSWORD
```
netsettler, line 31: I think probably we should prefer (untested):
```suggestion
        run: make publish
```
It does some additional checks, so you'll want to check that the checks it does are proper and will work on GA.

netsettler (approved):
My comment is not blocking.

PR #126, opened 2021-01-05
C4-503 Grab Environment API

willronchetti:
- Adds `get_beanstalk_environment_variables`, which will return information necessary to simulate any application given the caller has the appropriate access keys.
- Removes an `obsolete` tag from `create_db_snapshot`, which was set erroneously.

### DIFF
--- a/test/test_beanstalk_utils.py
+++ b/test/test_beanstalk_utils.py
@@ -362,7 +362,7 @@ def test_compute_prd_env_for_project():
             'OptionName': 'super_secret',
             'Value': 'i am secret'
          }
-     ], [('super_secret', 'i am secret')]),
+     ], {'super_secret': 'i am secret'}),
     ([
          {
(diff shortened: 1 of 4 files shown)


test/test_beanstalk_utils.py
```diff
@@ -352,3 +353,43 @@ def test_compute_prd_env_for_project():
...
+            'OptionName': 'something',
+            'Value': 'important'
+         },
+         {
+            'Namespace': 'something else',
+            'OptionName': 'not_secret',
+            'Value': 'i dont care about this value'
+         }
+     ], {})
+])
```
netsettler, line 390: Should probably have one more case that tests that you're accumulating two values when both are in `bs.ENV_VARIABLE_NAMESPACE`, but this is fine either way.

netsettler (approved):
This looks fine. My comment is not blocking.

PR #127, opened 2021-01-06
Beanstalk Bugfix

willronchetti:
- Parses Beanstalk API correctly and passes region.

### DIFF
--- a/dcicutils/beanstalk_utils.py
+++ b/dcicutils/beanstalk_utils.py
@@ -393,7 +393,7 @@ def _get_beanstalk_configuration_settings(env):
             }
     """
     try:
-        client = boto3.client('elasticbeanstalk', region_name='us-east-1')
+        client = boto3.client('elasticbeanstalk', region_name=REGION)
         config = client.describe_configuration_settings(ApplicationName='4dn-web', EnvironmentName=env)
         options = config['ConfigurationSettings'][0]['OptionSettings']  # guaranteed to be present
         return options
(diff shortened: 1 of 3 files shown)


dcicutils/beanstalk_utils.py
```diff
@@ -393,9 +393,9 @@ def _get_beanstalk_configuration_settings(env):
             }
     """
     try:
-        client = boto3.client('elasticbeanstalk')
+        client = boto3.client('elasticbeanstalk', region_name='us-east-1')
         config = client.describe_configuration_settings(ApplicationName='4dn-web', EnvironmentName=env)
-        options = config['ConfigurationSettings']['OptionSettings']  # guaranteed to be present
+        options = config['ConfigurationSettings'][0]['OptionSettings']  # guaranteed to be present
```
netsettler, line 398: If it's the only one, you could also do this as (untested):
```suggestion
        # These are guaranteed to be present
        [settings] = config['ConfigurationSettings']
        options = settings['OptionSettings']
```
but what you have is fine.

dcicutils/beanstalk_utils.py
```diff
@@ -393,9 +393,9 @@ def _get_beanstalk_configuration_settings(env):
             }
     """
     try:
-        client = boto3.client('elasticbeanstalk')
+        client = boto3.client('elasticbeanstalk', region_name='us-east-1')
```
netsettler, line 396: I've noticed this is sometimes not passed.

1. Should we have a wrapper that always adds this?
2. Should we have a variable that holds the region name? It _might_ actually want to be a configuration param.

A comment noting possible future work would probably be fine here.

netsettler (approved):
This is fine. My comments are not blocking.

PR #128, opened 2021-01-12
Changelog warnings (C4-511) and publish fixes (C4-512)

netsettler:
This fixes two primary issues:

* A missing changelog issue is an error, and @willronchetti wanted it downgraded to a warning.
* GitHub Actions publication was not using the script because of prompting. This amends the `scripts/publish` script to take a `--noconfirm` argument and sets up a `publish-for-ga` target for `make` so that we can use that to do publication.

Along the way I also did:

* Improve support in `MockFileSystem.open` for an `encoding=` argument. There was some missing data flow.
* To get some testing to not fail under GA, added support in `MockBotoS3Client` for `.put_object()` and `.list_objects()`.

### DIFF
--- a/.github/workflows/main-publish.yml
+++ b/.github/workflows/main-publish.yml
@@ -32,5 +32,5 @@ jobs:
           PYPI_USER: ${{ secrets.PYPI_USER }}
           PYPI_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
         run: | 
-          pip install poetry
+          make configure
           make publish-for-ga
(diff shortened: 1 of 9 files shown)


willronchetti (approved):

netsettler:
@willronchetti, the changes here were not passing tests on GA, so I've added additional substantive changes to fix that failing tests. Please review those. Thanks. 🙂

willronchetti (approved):
Approving the s3 test changes as well, which should reduce GA test flakiness

PR #129, opened 2021-01-14
Fix problematic pytest dependency (C4-521)

netsettler:
* Fix problem in 1.9.0 with unwanted dependency on
  `pytest.PytestConfigWarning` ([C4-521](https://hms-dbmi.atlassian.net/browse/C4-521)).
* Added some unit tests to run instead of integration tests for
  `s3_utils` in a number of cases.

### DIFF
--- a/test/test_s3_utils.py
+++ b/test/test_s3_utils.py
@@ -1,4 +1,5 @@
 import boto3
+import contextlib
 import datetime
 import io
 import pytest
@@ -9,6 +10,39 @@ from dcicutils.s3_utils import s3Utils
 from dcicutils.beanstalk_utils import compute_ff_prd_env, compute_cgap_prd_env, compute_cgap_stg_env
 from dcicutils.env_utils import get_standard_mirror_env, FF_PUBLIC_URL_STG, FF_PUBLIC_URL_PRD, CGAP_PUBLIC_URL_PRD
 from unittest import mock
+
+
+@contextlib.contextmanager
+def mocked_s3_integration(integrated_names, zip_suffix=""):
+    """
+    This does common setup of some mocks needed by zip testing.
+    """
+
+    zip_path_key = "zip_path" + zip_suffix
+    zip_filename_key = "zip_filename" + zip_suffix
+
+    b3 = MockBoto3()
+
+    ffenv = integrated_names['ffenv']
+
+    with mock.patch.object(s3_utils_module, "boto3", b3):
+
+        s3_connection = s3Utils(env=ffenv)
+
+        # Not needed when mocked.
+        # s3_connection.s3_delete_dir(prefix)
+
+        # In our mock, this won't exist already on S3 like in the integrated version of this test,
+        # so we have to pre-load to our mock S3 manually. -kmp 13-Jan-2021
+        s3_connection.s3.upload_file(Filename=integrated_names[zip_path_key],
+                                     Bucket=s3_connection.outfile_bucket,
+                                     Key=integrated_names[zip_filename_key])
+
+        s3_connection.s3.put_object(Bucket=s3_connection.outfile_bucket,
+                                    Key=integrated_names['filename'],
+                                    Body=str.encode('thisisatest'))
+
+        yield s3_connection
 
 
 @pytest.mark.parametrize('ff_ordinary_envname', ['fourfront-mastertest', 'fourfront-webdev', 'fourfront-hotseat'])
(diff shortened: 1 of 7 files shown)


test/test_s3_utils.py
```diff
@@ -11,13 +12,55 @@
...
     util = s3Utils(env=ff_ordinary_envname)
     assert util.sys_bucket == 'elasticbeanstalk-%s-system' % ff_ordinary_envname
 
 
+@pytest.mark.integrated
 def test_s3utils_creation_ff_stg():
+    # TODO: I'm not sure what this is testing, so it's hard to rewrite
+    #   But I fear this use of env 'staging' implies the GA test environment has overbroad privilege.
+    #   We should make this work without access to 'staging'.
+    #   -kmp 13-Jan-2021
```
willronchetti, line 63: It's just testing that s3utils resolves the s3 buckets correctly. It also tests resolving staging from `get_standard_mirror_env(compute_ff_prd_env())`.

test/test_s3_utils.py
```diff
@@ -43,7 +86,12 @@ def test_stg(ff_staging_envname):
     test_stg(stg_beanstalk_env)
 
 
+@pytest.mark.integrated
 def test_s3utils_creation_ff_prd():
+    # TODO: I'm not sure what this is testing, so it's hard to rewrite
+    #   But I fear this use of env 'data' implies the GA test environment has overbroad privilege.
+    #   We should make this work without access to 'data'.
+    #   -kmp 13-Jan-2021
```
willronchetti, line 94: Doing the exact same thing as the above test, except with production. The "creation" part here is the resolution of the S3 buckets.

test/test_s3_utils.py
```diff
@@ -69,13 +117,19 @@ def test_prd(ff_production_envname):
...
     util = s3Utils(env=cgap_ordinary_envname)
     assert util.sys_bucket == 'elasticbeanstalk-%s-system' % cgap_ordinary_envname
 
 
+@pytest.mark.integrated
 def test_s3utils_creation_cgap_prd():
+    # TODO: I'm not sure what this is testing, so it's hard to rewrite
+    #   But I fear this use of env 'data' implies the GA test environment has overbroad privilege.
+    #   We should make this work without access to 'data'.
+    #   -kmp 13-Jan-2021
```
willronchetti, line 132: Again, identical test to FF staging/production except with CGAP.

test/test_s3_utils.py
```diff
@@ -120,20 +176,35 @@ def test_s3utils_get_keys_for_data():
     assert keys_fs['server'] == keys['server']
 
 
+@pytest.mark.integrated
 def test_s3utils_get_keys_for_staging():
+    # TODO: I'm not sure what this is testing, so it's hard to rewrite
+    #   But I fear this use of env 'staging' implies the GA test environment has overbroad privilege.
+    #   We should make this work without access to 'staging'.
+    #   -kmp 13-Jan-2021
```
willronchetti, line 184: It has overboard privileges in ways that go well beyond this test - we might consider disabling this test on GA anyway in case it echos the `keys` dictionary, which will contain (temporary, refreshed daily) admin access keys.

test/test_s3_utils.py
```diff
@@ -120,20 +176,35 @@ def test_s3utils_get_keys_for_data():
...
     keys = util.get_ff_key()
     assert keys['server'] == 'http://staging.4dnucleome.org'
 
 
+@pytest.mark.integrated
 def test_s3utils_get_jupyterhub_key(basestring):
+    # TODO: I'm not sure what this is testing, so it's hard to rewrite
+    #   But I fear this use of env 'data' implies the GA test environment has overbroad privilege.
+    #   We should make this work without access to 'data'.
+    #   -kmp 13-Jan-2021
```
willronchetti, line 195: Same deal as above.

test/test_s3_utils.py
```diff
@@ -120,20 +176,35 @@ def test_s3utils_get_keys_for_data():
...
     assert key['server'] == 'https://jupyter.4dnucleome.org'
 
 
-def test_s3utils_get_higlass_key():
+@pytest.mark.integrated
+def test_s3utils_get_higlass_key_integrated():
+    # TODO: I'm not sure what this is testing, so it's hard to rewrite
+    #   But I fear this use of env 'staging' implies the GA test environment has overbroad privilege.
+    #   We should make this work without access to 'staging'.
+    #   -kmp 13-Jan-2021
```
willronchetti, line 207: Same deal as above.

willronchetti (approved):
This all looks good - I've tried to answer several questions about the tests in this PR as well.

test/test_s3_utils.py
netsettler: ok. it's probably possible to mock that. i was worried it was trying to test that particular files are in place on specific machines or that those files are in a particular form. that's something that can change dynamically and not in this code, and that mocking would not be able to test. I won't fix that for this PR but can work with it in the future.

PR #130, opened 2021-01-15
Fix bug that sometimes results in duplicated search results (C4-336)

netsettler:
Bug C4-336 details how sometimes `ff_utils.search_metadata`, by doing a series of Elastic Search calls that it pastes together into a single result, can return a list containing duplicated items. This fixes that.

(There is also a risk that it will return omitted items, but that is harder to fix and hopefully likely to be more rare in practice, since that happens mostly when information is removed rather than added, and that's a rarer event. But in any case, fixing at least some of this problem is better than fixing none of it.)

Things to check in the review:

* This creates a table during the search call fo the uuids. I'm assuming that the expected size of a result is such that this table is not likely to be overly big.
* This uses the `'uuid'` as the thing to look for to uniquely identify elements. I assume that's the right marker to check. This tries to be careful about the possibility of non-presence of that marker.

### DIFF
--- a/test/test_ff_utils.py
+++ b/test/test_ff_utils.py
@@ -826,6 +826,7 @@ def test_upsert_metadata(integrated_ff):
 def make_mocked_item(n):
     return {'uuid': str(n)}
 
+
 MOCKED_SEARCH_COUNT = 130
 MOCKED_SEARCH_ITEMS = [make_mocked_item(uuid_counter) for uuid_counter in range(MOCKED_SEARCH_COUNT)]
 
@@ -834,7 +835,7 @@ def constant_mocked_search_items():
     return MOCKED_SEARCH_ITEMS
 
 
-class InsertingMockedSearchItems():
+class InsertingMockedSearchItems:
     """
     This class is for use in simulating a situation in which an insertion occurs between calls to a search.
     The base set of data from which simulated results are taken will change on the second (POS=1) call,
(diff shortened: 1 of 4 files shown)


test/test_ff_utils.py
```diff
@@ -864,20 +926,38 @@ def mocked_search(url, auth, ff_env, retry_fxn):
...
 def test_search_metadata_integrated(integrated_ff, url):
     check_search_metadata(integrated_ff, url)
 
 
-def check_search_metadata(integrated_ff, url):
+def check_search_metadata(integrated_ff, url, expect_shortfall=False):
     if url != '':  # replace stub with actual url from integrated_ff
         url = integrated_ff['ff_key']['server'] + '/'
+
+    InsertingMockedSearchItems.reset()
```
willronchetti, line 960: It's not clear to me that this function call does anything in this test. There are no mocks active here, you are just triggering (repeatedly) a side-effect that is never realized? Am I missing something here?

willronchetti (approved):
Approving, but the comment I made I think needs addressing

aschroed (approved):

test/test_ff_utils.py
willronchetti: Ah, I needed to expand to see that there is an instance where the mock is active - but in the case of the integrated test it is dead code.

netsettler: It's a bit obscure. I added a comment.

PR #131, opened 2021-01-19
Misc functionality in service of C4-183

netsettler:
In `dcicutils.misc_utils`:

* New function `remove_element` to remove an element from a list (functionally rather than by side-effect).
* New class `TestApp` which is a synonym for `webtest.TestApp` but declared not to be a test case. (Classes with names that have "test" in their name are possible to confuse with test cases under `pytest`, so this has the necessary declaration to suppress that confusion.)
* Make `_VirtualAppHelper` use new `TestApp`.

### DIFF
--- a/dcicutils/misc_utils.py
+++ b/dcicutils/misc_utils.py
@@ -69,7 +69,23 @@ class VirtualAppError(Exception):
         return self.__repr__()
 
 
-class _VirtualAppHelper(webtest.TestApp):  # effectively disguises 'TestApp'
+class TestApp(webtest.TestApp):
+    """
+    Equivalent to webtest.TestApp, but pytest will not let the name confuse into thinking it's a test case.
+
+    A test case in PyTest is something that contains "test" in its name. We didn't pick the name TestApp,
+    but there may be tools that want to use TestApp for testing, and so this is a better place to inherit from,
+    since we've added an appropriate declaration to keep PyTest from confusing it with a TestCase.
+    """
+
+    __test__ = False  # This declaration asserts to PyTest that this is not a test case.
+
+
+class _VirtualAppHelper(TestApp):
+    """
+    A helper class equivalent to webtest.TestApp, except that it isn't intended for test use.
+    """
+
     pass
 
 
@@ -783,6 +799,27 @@ def check_true(test_value, message, error_class=None):
         error_class = RuntimeError
     if not test_value:
         raise error_class(message)
+
+
+def remove_element(elem, lst, raise_error=True):
+    """
+    Returns a copy of the given list with the first occurrence of the given element removed.
+
+    If the element doesn't occur in the list, an error is raised unless given raise_error=False,
+    in which case a copy of the original list is returned.
+
+    :param elem: an object
+    :param lst: a list
+    :param raise_error: a boolean (default True)
+    """
+
+    result = lst.copy()
+    try:
+        result.remove(elem)
+    except ValueError:
+        if raise_error:
+            raise
+    return result
 
 
 def remove_prefix(prefix, text, required=False):
--- a/dcicutils/misc_utils.py
+++ b/dcicutils/misc_utils.py
@@ -803,10 +803,10 @@ def check_true(test_value, message, error_class=None):
 
 def remove_element(elem, lst, raise_error=True):
     """
-    Returns a copy of the given list with the first occurrence of the given element removed.
+    Returns a shallow copy of the given list with the first occurrence of the given element removed.
 
     If the element doesn't occur in the list, an error is raised unless given raise_error=False,
-    in which case a copy of the original list is returned.
+    in which case a shallow copy of the original list is returned (with no elements removed).
 
     :param elem: an object
     :param lst: a list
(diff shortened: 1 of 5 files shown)


dcicutils/misc_utils.py
```diff
@@ -785,6 +801,27 @@ def check_true(test_value, message, error_class=None):
...
+
+    If the element doesn't occur in the list, an error is raised unless given raise_error=False,
+    in which case a copy of the original list is returned.
+
+    :param elem: an object
+    :param lst: a list
+    :param raise_error: a boolean (default True)
+    """
+
+    result = lst.copy()
```
willronchetti, line 816: I would put a comment here saying this is in fact a "true" copy, as this to me is typically read as a shallow copy (I know it isn't).

dcicutils/misc_utils.py
```diff
@@ -785,6 +801,27 @@ def check_true(test_value, message, error_class=None):
...
+    try:
+        result.remove(elem)
+    except ValueError:
+        if raise_error:
+            raise
```
willronchetti, line 821: If you're predicating on `raise_error`, shouldn't you catch blanket `Exception` as well?

test/test_misc_utils.py
```diff
@@ -1282,6 +1288,16 @@ def test_check_true():
...
 
+def test_remove_element():
+
+    assert remove_element('b', ['a', 'b', 'c', 'a', 'b', 'c']) == ['a', 'c', 'a', 'b', 'c']
+
+    with pytest.raises(ValueError):
+        assert remove_element('z', ['a', 'b', 'c', 'a', 'b', 'c'])
+
+    assert remove_element('z', ['a', 'b', 'c', 'a', 'b', 'c'], raise_error=False) == ['a', 'b', 'c', 'a', 'b', 'c']
+
```
willronchetti, line 1299: I would test the "copy" part of this ie:
```
lst = [1, 2, 3]
new_lst = remove_element(1, lst)
assert lst == [1, 2, 3]
assert new_lst == [2, 3]
```

test/test_misc_utils.py
```diff
@@ -1282,6 +1288,16 @@ def test_check_true():
     assert msg in str(e)
 
 
+def test_remove_element():
+
+    assert remove_element('b', ['a', 'b', 'c', 'a', 'b', 'c']) == ['a', 'c', 'a', 'b', 'c']
+
+    with pytest.raises(ValueError):
+        assert remove_element('z', ['a', 'b', 'c', 'a', 'b', 'c'])
```
willronchetti, line 1296: `assert` is not necessary here?

willronchetti (approved):
Approving with small, non-blocking comments

test/test_misc_utils.py
willronchetti: Might be worth having a version that removes from the source list as well (no copy).

dcicutils/misc_utils.py
netsettler: Oh, fascinating. In fact, I thought it a shallow copy. I'll have to investigate that.

dcicutils/misc_utils.py
netsettler: From the doc string: "If the element doesn't occur in the list, an error is raised unless given raise_error=False,
    in which case a copy of the original list is returned." You're saying you think the argument name implies more than the doc string? Some people might agree... I'm not sure if I do.

test/test_misc_utils.py
netsettler: You're right I should test the original list.

test/test_misc_utils.py
netsettler: Yeah...

dcicutils/misc_utils.py
netsettler: In fact list.copy() does shallow copying, which had been my expectation and was what I wanted, basically only enough copying to allow me to remove the item without affecting the original structure. I updated the doc string to say so.

dcicutils/misc_utils.py
netsettler: I'm going to leave it as is. If it throws other than `ValueError`, something else is wrong that probably should be flagged. That's the usual reason for being more specific than `Exception`.

test/test_misc_utils.py
netsettler: I had specific need of the non-side-effecting one.  I'm less certain about the side-effecting one, so I'm going to leave that until we have a need so I can be sure I understand the situation. Here's the test I'm now doing:
```
    old = ['a', 'b', 'c', 'a', 'b', 'c']
    new = remove_element('b', old)
    assert old is not new
    assert old == ['a', 'b', 'c', 'a', 'b', 'c']
    assert new == ['a', 'c', 'a', 'b', 'c']

    new = remove_element('z', old, raise_error=False)
    assert old is not new
    assert new == ['a', 'b', 'c', 'a', 'b', 'c']
    assert old == ['a', 'b', 'c', 'a', 'b', 'c']

    with pytest.raises(ValueError):
        remove_element('z', old)
    assert old == ['a', 'b', 'c', 'a', 'b', 'c']
```

test/test_misc_utils.py
netsettler: Fixed.

PR #135, opened 2021-03-22
ES snapshots as class and MockFileSystem ugprades

netsettler:
This PR has two primary functions

* It packages up the ES snapshot work I was doing in a consumable form that can be used selectively even while it's not fully ready.
* It makes some extensions to `MockFileSystem` that are helpful to fixing another bug in `cgap-portal` testing ([C4-636](https://hms-dbmi.atlassian.net/browse/C4-636)).

Also this bit of opportunism:

* It also makes a change to `misc_utils` to add `find_associations` and extend `find_association`, rescuing lost functionality that had been stranded on a branch that was never pursued.

Details:

* New `shapshot_utils` which exports two interesting quantities:

  * Class ``ElasticSearchDataCache`` useful for loading certain declared data and (on local machines only right now, and only when the environment variable `ENABLE_SNAPSHOTS` is `TRUE`). 

  * Other classes can build on this if defined with the `@es_data_cache` decorator.

* In `qa_utils`:

  * For `MockFileSystem`:

    * Initializing the class itself allows new keyword arguments `auto_mirror_files_for_read` and `do_not_auto_mirror`.

    * The `open` method, used for mocking, now supports modes involving `t` and `+`, such as `w+` or `rt`.

  * New context manager method ``mock_exists_open_remove`` that mocks these common functions (`os.path.exists`, `io.open`, and `os.remove`) for the mock file system that is its ``self``.

* In `misc_utils`:

  * Extend `find_association` to allow a predicate as a search value.

  * New function `find_associations` which is like `find_association` but returns a list of results, so doesn't err if more than one found.

### DIFF
--- a/dcicutils/misc_utils.py
+++ b/dcicutils/misc_utils.py
@@ -943,15 +943,41 @@ def count(seq, filter=None):
     return count_if(filter or identity, seq)
 
 
-def find_association(data, **kwargs):
+# Previous definition
+#
+# def find_association(data, **kwargs):
+#     for datum in data:
+#         mismatch = False
+#         for k, v in kwargs.items():
+#             if k in datum and datum[k] != v:
+#                 mismatch = True
+#         if not mismatch:
+#             return datum
+#     return None
+
+def find_associations(data, **kwargs):
+    found = []
     for datum in data:
         mismatch = False
         for k, v in kwargs.items():
-            if k in datum and datum[k] != v:
+            defaulted_val = datum.get(k)
+            if not (v(defaulted_val) if callable(v) else (v == defaulted_val)):
                 mismatch = True
+                break
         if not mismatch:
-            return datum
-    return None
+            found.append(datum)
+    return found
+
+
+def find_association(data, **kwargs):
+    results = find_associations(data, **kwargs)
+    n = len(results)
+    if n == 0:
+        return None
+    elif n == 1:
+        return results[0]
+    else:
+        raise ValueError("Got %s results when 1 was expected." % n)
 
 
 def keyword_as_title(keyword):
--- a/dcicutils/qa_utils.py
+++ b/dcicutils/qa_utils.py
@@ -444,13 +444,13 @@ class MockFileSystem:
     def open(self, file, mode='r', encoding=None):
         if FILE_SYSTEM_VERBOSE:
             print("Opening %r in mode %r." % (file, mode))
-        if mode == 'w':
+        if mode in ('w', 'wt', 'w+', 'w+t', 'wt+'):
             return self._open_for_write(file_system=self, file=file, binary=False, encoding=encoding)
-        elif mode == 'wb':
+        elif mode in ('wb', 'w+b', 'wb+'):
             return self._open_for_write(file_system=self, file=file, binary=True, encoding=encoding)
-        elif mode == 'r':
+        elif mode in ('r', 'rt', 'r+', 'r+t', 'rt+'):
             return self._open_for_read(file, binary=False, encoding=encoding)
-        elif mode == 'rb':
+        elif mode in ('rb', 'r+b', 'rb+'):
             return self._open_for_read(file, binary=True, encoding=encoding)
         else:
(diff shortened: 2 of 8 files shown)


dcicutils/qa_utils.py
```diff
@@ -430,32 +432,58 @@ def __exit__(self, exc_type, exc_val, exc_tb):
 class MockFileSystem:
     """Extremely low-tech mock file system."""
 
-    def __init__(self, files=None, default_encoding='utf-8'):
+    def __init__(self, files=None, default_encoding='utf-8', auto_mirror_files_for_read=False, do_not_auto_mirror=()):
```
willronchetti, line 435: I'd prefer `None` vs. `()` here but a minor detail.

dcicutils/snapshot_utils.py
```diff
@@ -0,0 +1,416 @@
...
+        for attr, value in attributes_and_values.items():
+            existing = getattr(cls, attr)
+            if existing and existing != value:
+                if cls.DEBUG_SNAPSHOTS:
+                    import pdb
+                    pdb.set_trace()
+                raise RuntimeError("Conflicting %s: %s (new) and %s (existing)." % (attr, value, existing))
+            setattr(cls, attr, value)
+
+    @classmethod
```
willronchetti, line 250: Can be `staticmethod` (is the case in multiple spots, won't mark them all)

dcicutils/snapshot_utils.py
```diff
@@ -0,0 +1,416 @@
...
+                    # PRINT("Deleting index files", index_names_string)
+                    PRINT("Deleting index files for prefix=", indexer_namespace)
+                result = es.indices.delete(index_names_string)
+                if cls.VERBOSE_SNAPSHOTS:
+                    ignorable(result)
+                    # PRINT("deletion result=", result)
+                    PRINT("Deleted index files for prefix=", indexer_namespace)
+            result = es.snapshot.restore(repository=repository_short_name,
+                                         snapshot=snapshot_name,
+                                         wait_for_completion=True)
```
willronchetti, line 357: Docstring would clarify the expected behavior here (you are wiping all prefixed indices prior to a restore - you don't necessarily need/want to do that).

Moreover docstring is missing on a few of these (not marking), but those that do have it are documented well.

willronchetti (approved):
These changes look good to me. Glad to have a start on the snapshot stuff. I think some things could be more clear but it looks to me like it should work. Feel free to adjust documentation/tests later on when we want to actually use this.

dcicutils/misc_utils.py
```diff
@@ -943,15 +943,41 @@ def count(seq, filter=None):
...
+
+def find_association(data, **kwargs):
+    results = find_associations(data, **kwargs)
+    n = len(results)
+    if n == 0:
+        return None
+    elif n == 1:
+        return results[0]
+    else:
+        raise ValueError("Got %s results when 1 was expected." % n)
```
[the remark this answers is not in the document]
netsettler: Yes, sorry, I always mean to put them there but it falls through the cracks sometimes because they're usually in service of something else I'm more focused on.  As it happens, this was already released (by accident - a tagging issue) so I'm just going to merge it as is. I'll follow later with doc strings in another PR.

dcicutils/qa_utils.py
netsettler: Yeah, a fair point.  In a later PR I'll fix that.  It was trying to be immutable and to default to a list, which this sort of does. But probably better to be None and have it replaced by a list in the body. In other languages (of course I mean Common Lisp) the initial value of the missing args are evaluated on each call so you can actually default things rationally. A pity Python didn't do this.

dcicutils/snapshot_utils.py
netsettler: It doesn't like to restore when there is stuff already there.

netsettler:
Because of the unusual situation of this already getting released to pypi, I'm going to just merge this as-is.  I'll follow up on some comments in another PR later.

PR #137, opened 2021-04-05
C4-617 Docker, ECR, ECS Utils

willronchetti:
- Adds 3 new modules with basic functionality needed for further development on the alpha stack
- TODO 1: add some more desired APIs?
- TODO 2: integrated test?

### DIFF
--- a/dcicutils/docker_utils.py
+++ b/dcicutils/docker_utils.py
@@ -9,7 +9,7 @@ class DockerUtils:
     def __init__(self):
         self.client = docker.from_env()
 
-    def login(self, ecr_repo_uri, ecr_user, ecr_pass):
+    def login(self, *, ecr_repo_uri, ecr_user, ecr_pass):
         """ Authenticates with ECR through Docker - takes as input the result of
             ecr_client.get_authorization_token()
         """
@@ -29,4 +29,5 @@ class DockerUtils:
         image.tag(ecr_repo_name, tag=tag)
 
     def push_image(self, tag, ecr_repo_name):
+        """ Pushes the tag - presumes the above tag_image method has been called. """
         return self.client.images.push(ecr_repo_name, tag=tag)
--- a/dcicutils/docker_utils.py
+++ b/dcicutils/docker_utils.py
@@ -18,8 +18,7 @@ class DockerUtils:
 
     def build_image(self, *, path: str, tag='latest', rm=True):
         """ Builds an image with the given tag on the given path. """
-        image, build_log = self.client.images.build(
-            path=path, tag=tag, rm=rm)
+        image, build_log = self.client.images.build(path=path, tag=tag, rm=rm)
         return image, build_log
 
     @staticmethod
--- a/dcicutils/docker_utils.py
+++ b/dcicutils/docker_utils.py
@@ -28,7 +28,8 @@ class DockerUtils:
         # works by side-effect
         image.tag(ecr_repo_name, tag=tag)
 
-    def push_image(self, tag, ecr_repo_name):
+    def push_image(self, *, tag, ecr_repo_name, auth_config):
         """ Pushes the tag - presumes the above tag_image method has been called. """
-        for line in self.client.images.push(ecr_repo_name, tag=tag, stream=True, decode=True):
+        for line in self.client.images.push(ecr_repo_name, tag=tag, stream=True, decode=True,
+                                            auth_config=auth_config):
             PRINT(line)
--- a/dcicutils/docker_utils.py
+++ b/dcicutils/docker_utils.py
@@ -1,4 +1,5 @@
 import docker
+from .misc_utils import PRINT
 
 
 class DockerUtils:
@@ -29,4 +30,5 @@ class DockerUtils:
 
     def push_image(self, tag, ecr_repo_name):
         """ Pushes the tag - presumes the above tag_image method has been called. """
-        return self.client.images.push(ecr_repo_name, tag=tag)
+        for line in self.client.images.push(ecr_repo_name, tag=tag, stream=True, decode=True):
+            PRINT(line)
--- a/dcicutils/docker_utils.py
+++ b/dcicutils/docker_utils.py
@@ -15,6 +15,8 @@ class DockerUtils:
             ecr_client.get_authorization_token()
         """
         # works by side effect
+        # XXX: not clear that the above^ is true, considering auth_config override is
+        # needed in 'push_images' - Will 4/6/21
         self.client.login(username=ecr_user, password=ecr_pass, registry=ecr_repo_uri)
 
     def build_image(self, *, path: str, tag='latest', rm=True):
--- a/dcicutils/docker_utils.py
+++ b/dcicutils/docker_utils.py
@@ -1,0 +1,32 @@
+import docker
+
+
+class DockerUtils:
+    """ Intended for use in conjunction with ECR Utils to facilitate automated image
+        building/uploading. The caller must have AWS perms and be running Docker.
+    """
+
+    def __init__(self):
+        self.client = docker.from_env()
+
+    def login(self, ecr_repo_uri, ecr_user, ecr_pass):
+        """ Authenticates with ECR through Docker - takes as input the result of
+            ecr_client.get_authorization_token()
+        """
+        # works by side effect
+        self.client.login(username=ecr_user, password=ecr_pass, registry=ecr_repo_uri)
+
+    def build_image(self, *, path: str, tag='latest', rm=True):
+        """ Builds an image with the given tag on the given path. """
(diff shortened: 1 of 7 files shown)


dcicutils/ecr_utils.py
```diff
@@ -0,0 +1,68 @@
...
+        NOTE 1: unlike s3Utils, the resolved resources (repository) will NOT be
+        created if it does not already exist.
+        NOTE 2: the (already created) ECR repository must have the env_name in the URI
+    """
+
+    def __init__(self, *, env_name='cgap-mastertest', local_repository='cgap-wsgi'):
+        """ Creates an ECR client on startup """
+        self.env = env_name
+        self.local_repository = local_repository
+        self.is_cgap = 'cgap' in env_name
```
netsettler, line 31: We may have a function for this in `env_utils`. If we don't, we should.

dcicutils/ecr_utils.py
```diff
@@ -0,0 +1,68 @@
...
+        self.client = boto3.client('ecr', region='us-east-1')  # XXX: constant?
+        self.url = None  # set by calling the below method
+
+    def resolve_repository_uri(self, url=None):
+        if not self.url or url:
+            PRINT('NOTE: Calling out to ECR')
+            try:
+                resp = self.client.describe_repositories()
+                for repo in resp.get('repositories', []):
+                    if self.env in repo['repositoryUri']:
```
netsettler, line 41: It was not obvious to me at first that this is a text search. I suspect it will match `fourfront-cgapdev` while searching for `fourfront-cgap`. I would like to see a real URL parser applied here or some more robust test. I worry this could cause some serious dyslexia between environments if not careful.

I regard this comment as blocking, though there are a number of ways to resolve it.

dcicutils/ecr_utils.py
```diff
@@ -0,0 +1,68 @@
...
+        """ Calls to boto3 to get authorization credentials for ECR.
+            Gives a result like this:
+                {
+                    'authorizationData': [{'authorizationToken': 'very long token....',
+                                            'expiresAt': <datetime>,
+                                            'proxyEndpoint': <ecr URL> ... }]
+                }
+        """
+        try:
+            return self.client.get_authorization_token()['authorizationData'][0]
```
netsettler, line 58: I think this should be written in the other style that will catch the case of more than one bit of data returned, which would mean the query was ambiguous. e.g., (untested):
```
[auth_datum] = self.client.get_authorization_token()['authorizationData']
return auth_datum
```
or something like that.

dcicutils/ecr_utils.py
```diff
@@ -0,0 +1,68 @@
...
+        created if it does not already exist.
+        NOTE 2: the (already created) ECR repository must have the env_name in the URI
+    """
+
+    def __init__(self, *, env_name='cgap-mastertest', local_repository='cgap-wsgi'):
+        """ Creates an ECR client on startup """
+        self.env = env_name
+        self.local_repository = local_repository
+        self.is_cgap = 'cgap' in env_name
+        self.client = boto3.client('ecr', region='us-east-1')  # XXX: constant?
```
netsettler, line 32: I think the region should probably be an argument to the `__init__` function, no? You could make it default from some class variable.

dcicutils/ecr_utils.py
```diff
@@ -0,0 +1,68 @@
...
+        except Exception as e:
+            PRINT('Could not acquire ECR authorization credentials: %s' % e)
+            raise
+
+    @staticmethod
+    def extract_ecr_password_from_authorization(*, authorization):
+        """ Extracts the password from the authorization token returned from the
+            above API call.
+        """
+        return base64.b64decode(authorization['authorizationToken']).replace(b'AWS:', b'').decode('utf-8')
```
netsettler, line 68: Comment documenting expected format? Is there really no chance that there could be more than one AWS in there? Are you sure you don't want to decode first, then use a prefix check? Can the `AWS:` occur anywhere in here?

test/test_ecs_utils.py
```diff
@@ -0,0 +1,32 @@
...
+from dcicutils.ecs_utils import ECSUtils
+
+
+@pytest.fixture(scope='module')
+def ecs_utils():
+    return ECSUtils(cluster_name='dummy-cluster')
+
+
+def mock_update_service(*, cluster, service, forceNewDeployment):
+    """ Mock matching the relevant API signature for below (we don't actually want
```
netsettler, line 12: Sometime we should discuss this doc string indentation style and why it bugs me. But it's OK for now.

test/test_ecs_utils.py
```diff
@@ -0,0 +1,32 @@
...
+        to trigger an ECS deploy in unit testing.
+    """
+    return
+
+
+@pytest.mark.parametrize('service_name', [
+    ECSUtils.WSGI, ECSUtils.INDEXER, ECSUtils.INGESTER
+])
+def test_ecs_utils_basic_should_proceed(ecs_utils, service_name):
+    with mock.patch.object(ecs_utils.client, 'update_service', mock_update_service):
```
netsettler, line 22: I might suggest making a `make_update_service_mock` that takes expected values as arguments and returns the function that is the mock. e.g.,
```
def make_update_service_mock(expected_cluster, expected_service, expected_forceNewDeployment):
    def mock_update_service(*, cluster, service, forceNewDeployment):
        assert cluster == expected_cluster
        assert service == expected_service
        assert forceNewDeployment == expected_forceNewDeployment
```
so you can later
```
with mock.patch.object(ecs_utils.client, 'update_service', make_mock_update_service(...)):
```
specifying expectations, etc.

test/test_ecr_utils.py
```diff
@@ -0,0 +1,44 @@
...
+        assert url == REPO_URL
+
+
+def test_ecr_utils_workflow():
+    """ Tests URL + Login via Docker_cli"""
+    ecr_cli = ECRUtils()
+    docker_cli = DockerUtils()
+    with mock.patch.object(ecr_cli.client, 'describe_repositories', mocked_describe_respositories):
+        ecr_cli.resolve_repository_uri()
+        ecr_user, ecr_pass = 'dummy', 'dummy'  # XXX: integrated test for this mechanism
```
netsettler, line 35: Minimally I might not pass the same thing for user and pass so you can later check they are passed in correct order. If these values were taken from environment variables, you could use the `override_environ` (I think it's in `qa_utils`, but if not it's in `misc_utils`) to bind the values. But that would

test/test_ecr_utils.py
```diff
@@ -0,0 +1,44 @@
+import pytest
+from unittest import mock
+from dcicutils.ecr_utils import ECRUtils
+from dcicutils.docker_utils import DockerUtils
+
+
+REPO_URL = '123456789.dkr.ecr.us-east-2.amazonaws.com/cgap-mastertest'
```
netsettler, line 7: I assume this is a sample and not a real one? Maybe a comment saying not to expect it to work in integration tests.

netsettler (approved):
See the comment on resolve_repository_uri in ecr_utils.py, which I regard as blocking.
My approval is conditional on doing something about that.

dcicutils/ecr_utils.py
willronchetti: Fixed with `if repo['repositoryUri'].endswith(self.env)`

dcicutils/docker_utils.py
```diff
@@ -28,6 +30,8 @@ def tag_image(*, image, tag, ecr_repo_name):
         # works by side-effect
         image.tag(ecr_repo_name, tag=tag)
 
-    def push_image(self, tag, ecr_repo_name):
+    def push_image(self, *, tag, ecr_repo_name, auth_config):
         """ Pushes the tag - presumes the above tag_image method has been called. """
-        return self.client.images.push(ecr_repo_name, tag=tag)
+        for line in self.client.images.push(ecr_repo_name, tag=tag, stream=True, decode=True,
+                                            auth_config=auth_config):
+            PRINT(line)
```
netsettler, line 37: Will the line be self-identifying (needing no prefix explaining why it's being typed out)?

dcicutils/ecr_utils.py
```diff
@@ -38,13 +42,19 @@ def resolve_repository_uri(self, url=None):
             try:
                 resp = self.client.describe_repositories()
                 for repo in resp.get('repositories', []):
-                    if self.env in repo['repositoryUri']:
-                        return repo['repositoryUri']
+                    if repo['repositoryUri'].endswith(self.env):
```
netsettler, line 45: Less worried about this failing, though theoretically `.endswith('/' + self.env)` is a little better. This is good enough.

netsettler (commented):
No blockers remain. This seems fine.

PR #138, opened 2021-04-06
JH Docker Mount Update

ebberg:
This correctly finds Open Data files in JH based on the goofys mounted filesystem in the Docker container.

### DIFF
--- a/dcicutils/jh_utils.py
+++ b/dcicutils/jh_utils.py
@@ -149,11 +149,20 @@ def find_valid_file_or_extra_file(obj_id, format):
     home_dir = '/home/jovyan'
     # hardcoded so that we can find which volume to look in by file type
     # will need to be updated if more File types are added to proc file bucket
-    if 'FileVistrack' in file_meta['@type'] or 'FileProcessed' in file_meta['@type']:
+    if file_meta['open_data_url']:
+        data_dir = 'open_data'
+        if 'FileVistrack' in file_meta['@type'] or 'FileProcessed' in file_meta['@type']:
+            file_type = 'wfoutput'
+        else:
+            file_type = 'files'
+        open_data_prefix = '{}'.format(file_type)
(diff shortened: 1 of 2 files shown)


dcicutils/jh_utils.py
```diff
@@ -149,11 +149,20 @@ def find_valid_file_or_extra_file(obj_id, format):
...
     # hardcoded so that we can find which volume to look in by file type
     # will need to be updated if more File types are added to proc file bucket
-    if 'FileVistrack' in file_meta['@type'] or 'FileProcessed' in file_meta['@type']:
+    if file_meta['open_data_url']:
+        data_dir = 'open_data'
+        if 'FileVistrack' in file_meta['@type'] or 'FileProcessed' in file_meta['@type']:
+            file_type = 'wfoutput'
+        else:
+            file_type = 'files'
+        open_data_prefix = '{}'.format(file_type)
```
willronchetti, line 158: Why not just call this `open_data_prefix` from the start? The string build here is unnecessary.

dcicutils/jh_utils.py
```diff
@@ -149,11 +149,20 @@ def find_valid_file_or_extra_file(obj_id, format):
     home_dir = '/home/jovyan'
     # hardcoded so that we can find which volume to look in by file type
     # will need to be updated if more File types are added to proc file bucket
-    if 'FileVistrack' in file_meta['@type'] or 'FileProcessed' in file_meta['@type']:
+    if file_meta['open_data_url']:
```
willronchetti, line 152: This will crash if this field isn't present. You need to use `.get('open_data_url', None)`.

willronchetti (changes requested):

dcicutils/jh_utils.py
ebberg: Changing this to be 'fourfront-webprod/{}', that was why it was separate.

dcicutils/jh_utils.py
ebberg: Updated.

PR #139, opened 2021-04-14
Add ES cluster resize capability

willronchetti:
- Adds `ElasticSearchServiceClient`, a wrapper for `boto3.client('es')`
- Implements `resize_elasticsearch_cluster`, issuing an update to the relevant settings
- Integrated test was performed on staging
- Unit tests mock the boto3 API

### DIFF
--- a/dcicutils/es_utils.py
+++ b/dcicutils/es_utils.py
@@ -26,8 +26,8 @@ class ElasticSearchServiceClient:
     DEFAULT_COLD_DATA_NODE_TYPE = 'c5.large.elasticsearch'
     DEFAULT_COLD_DATA_NODE_COUNT = 2
 
-    def __init__(self):
-        self.client = boto3.client('es')
+    def __init__(self, region_name='us-east-1'):
+        self.client = boto3.client('es', region_name=region_name)
 
     def resize_elasticsearch_cluster(self, *, domain_name, master_node_type, master_node_count,
                                      data_node_type, data_node_count=2):
--- a/dcicutils/es_utils.py
+++ b/dcicutils/es_utils.py
@@ -18,7 +18,7 @@ class ElasticSearchServiceClient:
      """
     DEFAULT_HOT_MASTER_NODE_TYPE = 'c5.large.elasticsearch'
     DEFAULT_HOT_MASTER_NODE_COUNT = 3
-    DEFAULT_HOT_DATA_NODE_TYPE = 'r5.2xlarge.elasticsearch'
+    DEFAULT_HOT_DATA_NODE_TYPE = 'c5.2xlarge.elasticsearch'
     DEFAULT_HOT_DATA_NODE_COUNT = 2
 
     DEFAULT_COLD_MASTER_NODE_TYPE = 't2.small.elasticsearch'  # does not matter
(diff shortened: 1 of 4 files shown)


dcicutils/es_utils.py
```diff
@@ -8,6 +10,64 @@
...
+     """
+    DEFAULT_HOT_MASTER_NODE_TYPE = 'c5.large.elasticsearch'
+    DEFAULT_HOT_MASTER_NODE_COUNT = 3
+    DEFAULT_HOT_DATA_NODE_TYPE = 'c5.2xlarge.elasticsearch'
+    DEFAULT_HOT_DATA_NODE_COUNT = 2
+
+    DEFAULT_COLD_MASTER_NODE_TYPE = 't2.small.elasticsearch'  # does not matter
+    DEFAULT_COLD_MASTER_NODE_COUNT = 0
+    DEFAULT_COLD_DATA_NODE_TYPE = 'c5.large.elasticsearch'
+    DEFAULT_COLD_DATA_NODE_COUNT = 2
```
netsettler, line 27: I don't see these values being used. Should they be the default for something?

test/test_es_utils.py
```diff
@@ -1,7 +1,95 @@
...
+            }
+        }
+
+    @staticmethod
+    def mock_update_es_bad_response(DomainName, ElasticsearchClusterConfig):
+        return {
+            'something_else': {
+                'blah': 403
+            }
+        }
```
netsettler, line 32: The real meat of this test is not just "does it return this?" but "is this function called with the right args?" so you need to tell it what args are the ones that should provoke this response.  (Same for the above mocks, I was too lazy to illustrate all of them.)

```suggestion
    @staticmethod
    def make_mock_update_es_bad_response(expected_domain, expected_config):
        def mock_update_es_bad_response(DomainName, ElasticsearchClusterConfig):
            assert expected_domain == DomainName
            assert expected_config == ElasticsearchClusterConfig
            return {
                'something_else': {
                    'blah': 403
                }
            }
```

test/test_es_utils.py
```diff
@@ -1,7 +1,95 @@
...
+                data_node_type='c5.large.elasticsearch',
+                data_node_count=2
+            )
+            assert not success
+
+    def test_elasticsearch_service_client_resize_bad_response(self):
+        """ Tests handling of a badly formatted response. """
+        client = ElasticSearchServiceClient()
+        with mock.patch.object(client.client, 'update_elasticsearch_domain_config',
+                               self.mock_update_es_bad_response):
```
netsettler, line 70: `To be useful, the following obviously-untested code would need to be filled out:

``suggestion
                               self.make_mock_update_es_bad_response(
                                  expected_domain='fourfront-newtest',  # <-- is that right?
                                  expected_config=...what should be here?...)):
```

netsettler (approved):
I don't see any really obvious problems here but I think the testing should be done differently to be meaningful.

netsettler (commented):
You also need to update the CHANGELOG.

dcicutils/es_utils.py
willronchetti: The idea is these could be used in foursight.

PR #141, opened 2021-05-06
Port Application Dockerization utils

willronchetti:
Adds some additional APIs needed for the ECS setup, including:
- `get_ecs_real_url`, which will extract the URL of the ECS WSGI Service
- `assume_identity`, used by the cgap-portal directly currently to pull environment configuration (see c4_519)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,10 +6,15 @@ dcicutils
 Change Log
 ----------
 
+1.16.1
+======
+
+** PR 141 Port Application Dockerization utils
+
 1.16.0
 ======
 
-**PR 141: Move override_environ and override_dict to misc_utils**
+**PR 142: Move override_environ and override_dict to misc_utils**
 
 * In ``misc_utils``:
 
--- a/dcicutils/ecs_utils.py
+++ b/dcicutils/ecs_utils.py
@@ -33,7 +33,7 @@ class ECSUtils:
         try:
             self.client.update_service(cluster=cluster_name, service=service_name,
                                        forceNewDeployment=True)
-            PRINT('Successfully updated ECS cluster %s service %s' % cluster_name, service_name)
+            PRINT('Successfully updated ECS cluster %s service %s' % (cluster_name, service_name))
         except Exception as e:
             PRINT('Error encountered triggering cluster update: %s' % e)
             raise
--- a/dcicutils/ecs_utils.py
+++ b/dcicutils/ecs_utils.py
@@ -9,11 +9,11 @@ class ECSUtils:
     """
 
     def __init__(self):
-        """ Cluster name in this case is the env name """
+        """ Creates a boto3 client for 'ecs'. """
         self.client = boto3.client('ecs', region_name=CGAP_ECS_REGION)  # same as ECR
 
     def list_ecs_clusters(self):
-        """ Returns a list of ECS clusters + some metadata. """
+        """ Returns a list of ECS clusters ARNs. """
         return self.client.list_clusters().get('clusterArns', [])
 
     def list_ecs_services(self, *, cluster_name):
--- a/dcicutils/ecs_utils.py
+++ b/dcicutils/ecs_utils.py
@@ -7,39 +7,39 @@ class ECSUtils:
     """ Utility class for interacting with ECS - mostly stubs at this point, but will likely
         expand a lot.
     """
-    WSGI = 'wsgi'  # XXX: might want to be joined with an identifier?
-    INDEXER = 'indexer'
-    INGESTER = 'ingester'
-    SERVICES = [
-        WSGI, INDEXER, INGESTER
-    ]
 
-    def __init__(self, *, cluster_name):
+    def __init__(self):
         """ Cluster name in this case is the env name """
-        self.cluster_name = cluster_name
         self.client = boto3.client('ecs', region_name=CGAP_ECS_REGION)  # same as ECR
 
-    def update_ecs_service(self, *, service_name=WSGI):
+    def list_ecs_clusters(self):
+        """ Returns a list of ECS clusters + some metadata. """
+        return self.client.list_clusters().get('clusterArns', [])
+
+    def list_ecs_services(self, *, cluster_name):
+        """ Returns a list of ECS Service names under the given cluster. There should be 4 of them as of right now -
+            adding more is considered a compatible change.
+        """
+        services = self.client.list_services(cluster=cluster_name).get('serviceArns', [])
+        if len(services) < 4:
+            raise Exception('Environment error! Expected at least 4 services - check that ECS finished orchestrating.')
+        return services
+
+    def update_ecs_service(self, *, cluster_name, service_name):
         """ Forces an update of this cluster's service, by default WSGI (for now)
             It is highly likely we want some variants of this - but will start here.
             Note that this just updates a particular service.
         """
-        if service_name not in self.SERVICES:
-            raise NotImplementedError('Specified service %s is an invalid service!' % service_name)
         try:
-            self.client.update_service(cluster=self.cluster_name, service=service_name,
+            self.client.update_service(cluster=cluster_name, service=service_name,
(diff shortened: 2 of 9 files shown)


dcicutils/ecs_utils.py
```diff
@@ -7,39 +7,39 @@ class ECSUtils:
...
-    ]
 
-    def __init__(self, *, cluster_name):
+    def __init__(self):
         """ Cluster name in this case is the env name """
-        self.cluster_name = cluster_name
         self.client = boto3.client('ecs', region_name=CGAP_ECS_REGION)  # same as ECR
 
-    def update_ecs_service(self, *, service_name=WSGI):
+    def list_ecs_clusters(self):
```
netsettler, line 15: This looks like an incompatible change. Needs a major version bump. See my comments on pyproject.toml.

CHANGELOG.rst
```diff
@@ -6,10 +6,15 @@ dcicutils
 Change Log
 ----------
 
+1.16.1
```
netsettler, line 9: Feels like this should be a major version bump. I guess if you want to declare there are no users of this yet, we could not do that. But then I would put it in doc somewhere that the modules you're playing with are experimental and not yet something for which we're bumping major versions on incompatible changes.

dcicutils/ecs_utils.py
```diff
@@ -7,39 +7,39 @@ class ECSUtils:
...
-            self.client.update_service(cluster=self.cluster_name, service=service_name,
+            self.client.update_service(cluster=cluster_name, service=service_name,
                                        forceNewDeployment=True)
+            PRINT('Successfully updated ECS cluster %s service %s' % cluster_name, service_name)
         except Exception as e:
             PRINT('Error encountered triggering cluster update: %s' % e)
             raise
 
-    def update_all_services(self):
+    def update_all_services(self, *, cluster_name):
```
netsettler, line 41: Incompatible change.

dcicutils/ecs_utils.py
```diff
@@ -7,39 +7,39 @@ class ECSUtils:
...
+    def list_ecs_services(self, *, cluster_name):
+        """ Returns a list of ECS Service names under the given cluster. There should be 4 of them as of right now -
+            adding more is considered a compatible change.
+        """
+        services = self.client.list_services(cluster=cluster_name).get('serviceArns', [])
+        if len(services) < 4:
+            raise Exception('Environment error! Expected at least 4 services - check that ECS finished orchestrating.')
+        return services
+
+    def update_ecs_service(self, *, cluster_name, service_name):
```
netsettler, line 28: Incompatible change.

dcicutils/secrets_utils.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def assume_identity():
+    """ Grabs application identity from the secrets manager.
+        Looks for environment variable IDENTITY, which should contain the name of
+        a secret in secretsmanager that is a JSON blob of core configuration information.
+        Default value is current value in the test account. This name should be the
+        name of the environment.
+    """
+    secret_name = os.environ.get('IDENTITY', 'dev/beanstalk/cgap-dev')
```
netsettler, line 15: Since we own the machine, I guess this is OK. I'm somewhat inclined to believe a better name like `AWS_IDENTITY` or even `AWS_SECRET_IDENTITY` is a better name for the env variable. A common name like `IDENTITY` may not be so secret. Clearly its protection is more how it's used and where it's available. But the name doesn't suggest the apparent semantics. I'm OK with just leaving it, but I wanted to say this out loud.

dcicutils/secrets_utils.py
```diff
@@ -0,0 +1,53 @@
...
+        service_name='secretsmanager',
+        region_name=region_name
+    )
+
+    try:
+        get_secret_value_response = client.get_secret_value(
+            SecretId=secret_name
+        )
+    except ClientError as e:  # leaving some useful debug info to help narrow issues
+        if e.response['Error']['Code'] == 'DecryptionFailureException':
```
netsettler, line 28: Are you planning to do more here? Or why not an `in` call and just one clause?

dcicutils/secrets_utils.py
```diff
@@ -0,0 +1,55 @@
...
+        service_name='secretsmanager',
+        region_name=region_name
+    )
+
+    try:
+        get_secret_value_response = client.get_secret_value(
+            SecretId=secret_name
+        )
+    except ClientError as e:  # leaving some useful debug info to help narrow issues
+        if e.response['Error']['Code'] in [
```
netsettler, line 37: At some point this list should perhaps become a variable, but the code is evolving so at your leisure.

test/test_cfn_utils.py
```diff
@@ -0,0 +1,53 @@
+import pytest
+from unittest import mock
+from dcicutils.cloudformation_utils import get_ecs_real_url
+
+
+class MockedCfnClient():
```
netsettler, line 6: Maybe something to generalize sometime for `qa_utils` but good enough for now.

test/test_ecs_utils.py
```diff
@@ -6,7 +6,24 @@
...
+        'serviceArns':
+            [
+                'arn:aws:ecs:us-east-2:1234566777:service/c4-ecs-trial-alpha-stack-CGAPDockerCluster-Z4m1uYa2J11O/c4-ecs-trial-alpha-stack-CGAPIndexerService-YihcMquIc354',
+                'arn:aws:ecs:us-east-2:1234566777:service/c4-ecs-trial-alpha-stack-CGAPDockerCluster-Z4m1uYa2J11O/c4-ecs-trial-alpha-stack-CGAPDeploymentService-NRyGGBTSnqbQ',
+                'arn:aws:ecs:us-east-2:1234566777:service/c4-ecs-trial-alpha-stack-CGAPDockerCluster-Z4m1uYa2J11O/c4-ecs-trial-alpha-stack-CGAPIngesterService-QRcdjlE5ZJS1',
+                'arn:aws:ecs:us-east-2:1234566777:service/c4-ecs-trial-alpha-stack-CGAPDockerCluster-Z4m1uYa2J11O/c4-ecs-trial-alpha-stack-CGAPWSGIService-oDZbeVVWjZMq'
+            ],
+        'ResponseMetadata':
+            {'RequestId': 'not-a-uuid',
+             'HTTPStatusCode': 200}
```
netsettler, line 25: What is this telling me? The request id looks like an error but it's returning a 200. I'm not sure I fully get the context though.

test/test_ecs_utils.py
```diff
@@ -17,17 +34,8 @@ def mock_update_service(*, cluster, service, forceNewDeployment):  # noQA - AWS
...
-@pytest.mark.parametrize('service_name', [
-    'badservicename', 5, None
-])
-def test_ecs_utils_basic_should_fail(ecs_utils, service_name):
-    with pytest.raises(Exception):
-        ecs_utils.update_ecs_service(service_name=service_name)
+def test_ecs_utils_will_update_services(ecs_utils):
+    """ Tests basic code interaction, no integrated testing right now. """
+    with mock.patch.object(ecs_utils.client, 'list_services', mock_list_ecs_services):
+        with mock.patch.object(ecs_utils.client, 'update_service', mock_update_service):
```
netsettler, line 40: If you just don't supply a mock, and instead pick it up using `as`, it will give you a magic mock that you can then ask whether it got called using .assert_called_with(...) to refer to the most recent call or if there are a lot there is a way to get all the calls. e.g., from https://docs.python.org/3/library/unittest.mock.html you can see:
```
mock = Mock(return_value=None)
mock()
mock(3, 4)
mock(key='fish', next='w00t!')
mock.call_args_list
[call(), call(3, 4), call(key='fish', next='w00t!')]
expected = [(), ((3, 4),), ({'key': 'fish', 'next': 'w00t!'},)]
mock.call_args_list == expected
True
```

netsettler (approved):
Seems OK to me. I didn't see any blockers.

PR #142, opened 2021-05-12
S3 utils can be initialized with an S3 bucket C4-554

ebberg:
If an environment variable, GLOBAL_ENV_BUCKET, is set to an S3 Bucket, s3Utils will attempt to access this bucket when initialized, and look for a configuration file there. It will then use this config to set the names of the S3 buckets it uses for a ff environment.

### DIFF
--- a/dcicutils/cloudformation_utils.py
+++ b/dcicutils/cloudformation_utils.py
@@ -10,8 +10,24 @@ def get_ecs_real_url(env):
     cfn_client = boto3.client('cloudformation')
     stacks = cfn_client.describe_stacks().get('Stacks', [])
     for stack in stacks:
-        for output in stack['Outputs']:
+        for output in stack.get('Outputs', []):
             if output.get('OutputKey', '') == ('ECSApplicationURL%s' % env.replace('-', '')):
                 return output.get('OutputValue')
+    PRINT('Did not locate the server from Cloudformation! Check ECS Stack metadata.')
+    return ''
+
+
+def get_ecr_repo_url(env):
+    """ Gets the ECR repository URL of the repository from which ECS pulls application
+        image runtimes from.
+    """
+    cfn_client = boto3.client('cloudformation')
+    stacks = cfn_client.describe_stacks().get('Stacks', [])
+    for stack in stacks:
+        for output in stack.get('Outputs', []):
+            if output.get('OutputKey', '') == 'C4EcrTrialAlphaECRRepoURL':  # XXX: hardcoded constant from 4dn-cloud-infra
+                val = output.get('OutputValue')
+                if env in val:
+                    return val
     PRINT('Did not locate the server from Cloudformation! Check ECS Stack metadata.')
     return ''
--- a/dcicutils/s3_utils.py
+++ b/dcicutils/s3_utils.py
@@ -3,6 +3,7 @@ import json
 import boto3
 import os
 import mimetypes
+import urllib.request
 from zipfile import ZipFile
 from io import BytesIO
 import logging
@@ -66,6 +67,16 @@ class s3Utils(object):  # NOQA - This class name violates style rules, but a lot
             env_config = json.loads(get_response['Body'].read())
             return env_config
 
+    @staticmethod
+    def fetch_health_page_json(url, use_urllib):
+        if use_urllib is False:
+            return requests.get(url).json()
+        else:
+            res = urllib.request.urlopen(url)
+            res_body = res.read()
+            j = json.loads(res_body.decode("utf-8"))
+            return j
+
     def __init__(self, outfile_bucket=None, sys_bucket=None, raw_file_bucket=None,
                  blob_bucket=None, metadata_bucket=None, env=None):
         """ Initializes s3 utils in one of three ways:
@@ -85,7 +96,7 @@ class s3Utils(object):  # NOQA - This class name violates style rules, but a lot
             ff_url = env_config['fourfront']
             health_json_url = '{ff_url}/health?format=json'.format(ff_url=ff_url)
             logger.warning('health json url: {}'.format(health_json_url))
-            health_json = requests.get(health_json_url).json()
+            health_json = self.fetch_health_page_json(url=health_json_url, use_urllib=True)
             sys_bucket = health_json['system_bucket']
             outfile_bucket = health_json['processed_file_bucket']
             raw_file_bucket = health_json['file_upload_bucket']
--- a/dcicutils/s3_utils.py
+++ b/dcicutils/s3_utils.py
@@ -24,16 +24,68 @@ class s3Utils(object):  # NOQA - This class name violates style rules, but a lot
     RAW_BUCKET_TEMPLATE = "elasticbeanstalk-%s-files"
     BLOB_BUCKET_TEMPLATE = "elasticbeanstalk-%s-blobs"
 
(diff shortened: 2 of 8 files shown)


dcicutils/s3_utils.py
```diff
@@ -24,16 +25,70 @@ class s3Utils(object):  # NOQA - This class name violates style rules, but a lot
...
+        head_response = s3_client.head_bucket(Bucket=global_bucket)
+        status = head_response['ResponseMetadata']['HTTPStatusCode']  # should be 200; raise error for 404 or 403
+        if status != 200:
+            raise Exception('Could not access GLOBAL_BUCKET_ENV {global_bucket}: status: {status}'.format(
+                global_bucket=global_bucket, status=status))
+        # list contents of global bucket, look for a match with the global bucket name
+        list_response = s3_client.list_objects_v2(Bucket=global_bucket)
+        # no match, raise exception
+        if list_response['KeyCount'] < 1:
+            raise Exception('No config objects found in global bucket {global_bucket}'.format(
```
netsettler, line 43: Same issue here again, although this suggests the code I wrote for GlobalBucketAccessFailure above is not general enough, so I could rewrite it:

```
class GlobalBucketException(S3UtilsException):
    MESSAGE = "Problem with"
    def __init__(self, **kwargs):
        for k, v in kwargs.items():
            setattr(self, k, v)
        super().__init__(self.MESSAGE.format(**kwargs))

class GlobalBucketAccessFailure(GlobalBucketException):
    MESSAGE = "Could not access GLOBAL_BUCKET_ENV {bucket}: status: {status}"
    def __init__(self, bucket, status):
        super().__init__(bucket=bucket, status=status)
```
so that we can add
```
class GlobalBucketMissingConfig(GlobalBucketException):
    MESSAGE = "No config objects found in global bucket {bucket}."
    def __init__(self, bucket):
        super().__init__(bucket=bucket)
```
and we can write (untested):
```
    raise GlobalBucketMissingConfig(bucket=global_bucket)
```

dcicutils/s3_utils.py
```diff
@@ -24,16 +25,70 @@ class s3Utils(object):  # NOQA - This class name violates style rules, but a lot
...
+                global_bucket=global_bucket))
+        keys = [list_response['Contents'][i]['Key'] for i in range(0, len(list_response['Contents']))]
+        config_filename = None
+        for filename in keys:
+            # multiple matches, raise exception
+            if filename in global_bucket and config_filename is not None:
+                raise Exception('multiple matches for global env bucket: {global_bucket}; keys: {keys}'.format(
```
netsettler, line 50: Here again (untested, and the last one I'll spell out):
```
class GlobalBucketMultipleMatches(GlobalBucketException):
    MESSAGE = "Multiple matches for global env bucket: {global_bucket}; keys: {keys}"
    def __init__(self, bucket, keys):
        super().__init__(bucket=bucket, keys=keys)
```
so that here in context we can do (yup, still untested):
```
    raise GlobalBucketMultipleMatches(bucket=global_bucket, keys=keys)
```
and so on for the other exceptions below.

dcicutils/s3_utils.py
```diff
@@ -24,16 +25,70 @@ class s3Utils(object):  # NOQA - This class name violates style rules, but a lot
...
+                    global_bucket=global_bucket,
+                    keys=keys,
+                ))
+            elif filename in global_bucket:
+                config_filename = filename
+            else:
+                pass
+        if not config_filename:
+            raise Exception('no matches for global env bucket: {global_bucket}; keys: {keys}'.format(
```
netsettler, line 59: See above.

dcicutils/s3_utils.py
```diff
@@ -24,16 +25,70 @@ class s3Utils(object):  # NOQA - This class name violates style rules, but a lot
...
+    def __init__(self, outfile_bucket=None, sys_bucket=None, raw_file_bucket=None,
+                 blob_bucket=None, metadata_bucket=None, env=None):
+        """ Initializes s3 utils in one of three ways:
+        1) If 'GLOBAL_BUCKET_ENV' is set to an S3 env bucket, use that bucket to fetch the env for the buckets.
+           We then use this env to build the bucket names.
+        2) With no global env set, if we instead pass in the env kwarg, we use this kwarg to build the bucket names.
+        3) With no global env or env kwarg, we expect bucket kwargs to be set, and use those as bucket names directly.
         """
         # avoid circular ref
         from .beanstalk_utils import get_beanstalk_real_url
```
netsettler, line 78: The need for this import is painful. We should really try to sort this out. I'll see if I  can help do that later. OK for now. Please adjust the comment, though:
```suggestion
        # TODO: The import here avoids circular ref, but we really need to fix that. -kmp 17-May-2021
        from .beanstalk_utils import get_beanstalk_real_url
```

poetry.lock
```diff
@@ -8,17 +8,17 @@ python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
 
 [[package]]
 name = "attrs"
-version = "20.3.0"
+version = "21.2.0"
```
netsettler, line 11: A lot of things being updated here. Did we mean to take all these updates?

dcicutils/s3_utils.py
```diff
@@ -24,16 +25,70 @@ class s3Utils(object):  # NOQA - This class name violates style rules, but a lot
...
         self.url = ''
         self.s3 = boto3.client('s3', region_name='us-east-1')
-        if sys_bucket is None:
+        global_bucket = os.environ.get('GLOBAL_BUCKET_ENV')
+        if global_bucket:
```
netsettler, line 82: There is no `else` for this `if`. Is that OK? 

Also, some explanation somewhere of what a "global bucket env" is would be good.

netsettler (approved):
I don't see any blocking problems, though I did make some comments I hope are worth considering if there's time. (They could be attended to another time. There's no urgency to them.)

netsettler (commented):
Actually, I guess I want to add that some testing of this new functionality is in order I don't see test cases that talk at all about the new functionality.

dcicutils/cloudformation_utils.py
```diff
@@ -10,8 +10,24 @@ def get_ecs_real_url(env):
...
+
+def get_ecr_repo_url(env):
+    """ Gets the ECR repository URL of the repository from which ECS pulls application
+        image runtimes from.
+    """
+    cfn_client = boto3.client('cloudformation')
+    stacks = cfn_client.describe_stacks().get('Stacks', [])
+    for stack in stacks:
+        for output in stack.get('Outputs', []):
+            if output.get('OutputKey', '') == 'C4EcrTrialAlphaECRRepoURL':  # XXX: hardcoded constant from 4dn-cloud-infra
```
netsettler, line 28: Hardcoded constants should have names.

dcicutils/cloudformation_utils.py
```diff
@@ -10,8 +10,24 @@ def get_ecs_real_url(env):
...
+                val = output.get('OutputValue')
+                if env in val:
+                    return val
+    PRINT('Did not locate the server from Cloudformation! Check ECS Stack metadata.')
```
netsettler, line 32: Log perhaps? Printing is good if you're doing something interactively. Is that a reliable assumption here?

PR #144, opened 2021-05-25
Add known_bug_expected and related support

netsettler:
Some stuff I had laying around in a sandbox I'm trying to clear out...

* In `dcicutils.qa_utils`:

  * Add `known_bug_expected` to mark situations in testing where a named bug is expected (one for which there is a JIRA ticket), allowing managing of the error handling by setting the bug's status as `fixed=False` (the default) or `fixed=True`.

* In `dcicutils.misc_utils`:

  * Add `capitalize1` to uppercase the first letter of something, leaving other case alone (rather than forcing it lower).

* In new `dcicutils.exceptions` to hold onto exception classes, with error classes used by `known_bug_expected`:
  * `KnownBugError`
  * `UnfixedBugError`
  * `WrongErrorSeen`
  * `ExpectedErrorNotSeen`
  * `FixedBugError`
  * `WrongErrorSeenAfterFix`
  * `UnexpectedErrorAfterFix`

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,25 @@ dcicutils
 Change Log
 ----------
 
+
+1.17.0
+======
+
+**PR 142: Add known_bug_accepted and related support**
+
+* In ``misc_utils``:
+
+  * Add ``capitalize1`` to uppercase the first letter of something,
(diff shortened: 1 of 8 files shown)


willronchetti (approved):
This all looks fairly reasonable to me

PR #145, opened 2021-06-10
Fix internal import problems (C4-663)

netsettler:
This fixes some internal import problems that create an unwanted dependency on pytest.

It also moves `local_attrs` into the `misc_utils` so it can be used by other libraries without a `pytest` dependency.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,12 +6,23 @@ dcicutils
 Change Log
 ----------
 
+
+1.18.1
+======
+
+**PR 145 Fix internal import problems**
+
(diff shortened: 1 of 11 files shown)


drio18 (approved):
Looks good to me. One thing I'll note is that you're moving from absolute to relative imports, while I've noticed in cgap-portal we've been moving in the opposite direction.

netsettler:
I'll discuss that elsewhere, but as of Python 3, but relative imports are always preferred where possible. Someone (or some tool) may have been sloppy in cgap-portal. We can explore that under separate cover, though.

PR #149, opened 2021-07-28
Configurable env_utils for orchestrated environments (C4-689)

netsettler:
```
THIS PR REQUIRES A TRANSITION STRATEGY (see note at end) AND BETA TESTING.
CONSEQUENTLY, THIS DOES NOT NEED DETAILED REVIEW AT THIS TIME.
```

This PR is responsive to [Need a configurable env_utils for orchestrated applications in isolated accounts (C4-689)](https://hms-dbmi.atlassian.net/browse/C4-689).

This does several things:

* Renames `dcicutils/env_utils.py` to `dcicutils/env_utils_legacy.py` so that old functionality is not disturbed.

* Introduces new file `dcicutils/env_utils.py` that wraps the legacy support, introducing a new configurable set of functions.

* **(incompatible change)** Removes support for any variables (other than those containing functions) in `dcicutils.env_utils`. The variables in what was formerly `dcicutils.env_utils` and what is now `dcicutils.env_utils_legacy` are not able to be represented in this way, and really should not be anyway. They would need two (or more) different values per variable, and the old values would only apply in the old environment.

* Does a major version bump to accommodate the incompatible change.

**NOTE:** Although strictly incompatible, it's possible and desirable to phase more gradually by a compatible migration path. Imports of these variables in other repos be found and transitioned to function calls, introducing new function calls as needed. See separate ticket [Rewrite uses of variables imported from dcicutils.env_utils (C4-690)](https://hms-dbmi.atlassian.net/browse/C4-690).


dcicutils/env_utils.py
```diff
@@ -1,318 +1,448 @@
...
+
+
+e = EnvNames
+p = PublicUrlParts
+c = ClassificationParts
+
+_MISSING = object()
+
+
+class EnvUtils:
```
willronchetti, line 110: This whole thing needs to be documented more I'd say.

netsettler:
This work continues on branch `kmp_alt2_env_utils`.

PR #150, opened 2021-08-01
Add json_leaf_subst, conjoined_list and disjoined_list

netsettler:
* In `misc_utils`, add `json_leaf_subst` to do substitutions at the leaves (atomic parts) of a JSON object.

* In `lang_utils`, add `conjoined_list` and `disjoined_list` to get a comma-separated list in ordinary English form with an "and" or an "or" before the last element.

* In `lang_utils`, using new `conjoined_list` and `disjoined_list`, add higher level operations `there_are` and `must_be_one_of` to construct proper English wordings for certain common kinds of expressions.

* In `exceptions`, using new `must_be_one_of`, add `InvalidParameterError`.

* In `deployment_utils`, added support for initializing the `identity` and `tibanna_logs_bucket` from an `ENCODED_IDENTITY` and `ENCODED_TIBANNA_LOGS_BUCKET` variable or `--identity` and `--tibanna_logs_bucket` command line arg in the ini-maker.

* In `s3_utils`, better initialization of `.metadata_bucket` and addition of new `.tibanna_logs_bucket`.


willronchetti (approved):

dcicutils/s3_utils.py
```diff
@@ -29,6 +29,8 @@ class s3Utils(object):  # NOQA - This class name violates style rules, but a lot
     OUTFILE_BUCKET_TEMPLATE = "elasticbeanstalk-%s-wfoutput"
     RAW_BUCKET_TEMPLATE = "elasticbeanstalk-%s-files"
     BLOB_BUCKET_TEMPLATE = "elasticbeanstalk-%s-blobs"
+    METADATA_BUCKET_TEMPLATE = "elasticbeanstalk-%s-metadata-bundles"
+    TIBANNA_LOGS_BUCKET_TEMPLATE = "elasticbeanstalk-%s-tibanna-logs"
```
willronchetti, line 33: I would not add this as it isn't correct in the old account (hardcoded to `tibanna-logs`) or the new account.

dcicutils/s3_utils.py
```diff
@@ -148,6 +158,8 @@ def __init__(self, outfile_bucket=None, sys_bucket=None, raw_file_bucket=None,
                 outfile_bucket = self.OUTFILE_BUCKET_TEMPLATE % env
                 raw_file_bucket = self.RAW_BUCKET_TEMPLATE % env
                 blob_bucket = self.BLOB_BUCKET_TEMPLATE % env
+                metadata_bucket = self.METADATA_BUCKET_TEMPLATE % env
+                tibanna_logs_bucket = self.TIBANNA_LOGS_BUCKET_TEMPLATE % env
```
willronchetti, line 162: I would leave this as `None` for now, or simply `tibanna-logs`.

test/ini_files/cg_any.ini
```diff
@@ -9,6 +9,8 @@ blob_bucket = elasticbeanstalk-${S3_BUCKET_ENV}-blobs
 system_bucket = elasticbeanstalk-${S3_BUCKET_ENV}-system
 metadata_bundles_bucket = elasticbeanstalk-${S3_BUCKET_ENV}-metadata-bundles
 sentry_dsn = ${SENTRY_DSN}
+identity = ${IDENTITY}
+tibanna_logs_bucket = elasticbeanstalk-${S3_BUCKET_ENV}-tibanna-logs
```
willronchetti, line 13: Same issue here - this bucket is always `tibanna-logs` in the current account. We should maybe discuss how we want to handle this.

test/ini_files/cgap.ini
```diff
@@ -9,6 +9,7 @@ blob_bucket = elasticbeanstalk-fourfront-cgap-blobs
 system_bucket = elasticbeanstalk-fourfront-cgap-system
 metadata_bundles_bucket = elasticbeanstalk-fourfront-cgap-metadata-bundles
 sentry_dsn = ${SENTRY_DSN}
+tibanna_logs_bucket = elasticbeanstalk-fourfront-cgap-tibanna-logs
```
willronchetti, line 12: Doing this won't break anything as nothing uses the portal as an authoritative source on this bucket, but it should be noted as in other places this bucket name is wrong. This comment applies to all the legacy ini files.

dcicutils/s3_utils.py
netsettler: It should default to what works in the legacy account. It doesn't have to work for alpha. So I think you're saying I should just have it be "tibanna-logs".

dcicutils/s3_utils.py
netsettler: Well, if that variable value was set correctly, it would work this way.

test/ini_files/cg_any.ini
netsettler: It is perhaps not obvious, but this file only works in legacy cgap. There is a separate alpha any, and it will pick up the value of a variable that will be reliably set. This is only accommodating old code.

PR #153, opened 2021-08-30
Miscellaneous orchestration support

netsettler:
This PR does three main things, each quite different, but each supporting some aspect of orchestration.

* Adds support for dynamically writing shell scripts that is useful for scripts used in the tibanna setup of 4dn-cloud-infra.
* Adds support for more modules to get auto-doc at ReadTheDocs, and adds unit tests that try to help that stay up-to-date. The desire to refer people to the `env_utils` and `s3_utils` documentation made this an issue. Only a handful of the `dcicutils` modules are autodoc'd right now.
* Adds some enhanced pluralization support for multi-word combinations (like "file to load" becoming "files to load")  so that various  other high-level operations like `n_of` and `there_are` can takes such expressions as noun form arguments and do the right thing.

See the `CHANGELOG.rst` for a more detailed summary of changes.

There are a number of small changes due to `flake8` fussing at me. I set up a `.flake8` rule to ignore `F541`, which fusses uselessly about using `f"..."` with no expression placeholders. I don't see any reason to care about that. Weird thing about `flake8` is that doing so overrides their default list of options, so the list has to specifically enumerate the options you think they are defaulting and won't pick up new ones later. :(

### DIFF
--- a/dcicutils/lang_utils.py
+++ b/dcicutils/lang_utils.py
@@ -68,9 +68,7 @@ class EnglishUtils:
 
     _NOUN_WITH_PREPOSITIONAL_ATTACHMENT = re.compile(
         # Note use of *? to get minimal match in group1, so that we'll find the first preposition, not a later one
-        # f"^(?:([a-z]*?)([-])+({'|'.join(COMMON_PREPOSITIONS)})([-]+)([a-z-]*)|([a-z- ]*?[a-z-]?)([ ]+)({'|'.join(COMMON_PREPOSITIONS)})([ ]+)(.*))$",
-
-        # The use of "*?" makes us treat multiple prepositions like a-of-b-of-c by recursing as
+        # Specifically, the use of "*?" makes us treat multiple prepositions like a-of-b-of-c by recursing as
         # (a)(-)(of)(-)(b-of-c). If we used "*" instead of "*?", we would get (a-of-b)(-)(of)(-)(c).
         # If we were only doing hyphenated items, it wouldn't matter a whole lot.
         # But in the words part, it matters a great deal because with spaces and hyphens interleaved, there is a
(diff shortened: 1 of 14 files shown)


dcicutils/lang_utils.py
```diff
@@ -50,18 +52,86 @@ def _special_case_plural(cls, word: str) -> str:
...
+                 # or these 5 will match (and the previous 5 will not)...
+                 # to match shortest (i.e., first) words (included hyphenated words) leading to " <prep> "
+                 # so 'the son-in-law of the proband' becomes (the son-in-law)( )(of)( )(the proband)
+                 ([a-z]        # first token starts with an alphabetic,
+                  [a-z- ]*?    # matches any number of words, which may be hyphenated,
+                  [a-z])       # and ends in an alphabetic
+                 ([ ]+)        # unlike with hyphenation, any number of pre-<prep> spaces is ok
+                 ({'|'.join(_COMPOUND_PLURAL_PREPOSITIONS).replace(' ', '[ ]')}) # matches the prep, as (about|at|...)
+                 ([ ]+)        # any amount of whitespace on the other side
+                 (.*)          # anything else that follows first space-delimeted preposition
```
willronchetti, line 95: Nice little regex refresher 😄   - I am taking what you say here at face value and not double checking it, just FYI as this is not one of my stronger areas.

willronchetti (approved):

dcicutils/lang_utils.py
netsettler: the general approach is something i've been refining for years in various environments so i have a general theory of what i'm trying to do, but i'm doing the implementation from scratch here just from the ideas, so some risk of error. And I've never tried to do it with regexps.

still, the risk of being wrong is not high. the unit test coverage is decent, but i could probably fix bugs if they came up.

PR #154, opened 2021-08-31
s3_utils EnvManager

netsettler:
See the `CHANGELOG.rst` for details


dcicutils/s3_utils.py
```diff
@@ -55,49 +57,13 @@ class s3Utils(object):  # NOQA - This class name violates style rules, but a lot
...
-            # we found a match, so fetch that file as config
-            get_response = s3_client.get_object(Bucket=global_bucket, Key=config_filename)
-            env_config = json.loads(get_response['Body'].read())
-            return env_config
+        return EnvManager.verify_and_get_env_config(s3_client=s3_client,
+                                                    global_bucket=global_bucket,
+                                                    env=env)
 
-    @staticmethod
+    @staticmethod  # backwawrd compatibility in case other repositories are using this
```
willronchetti, line 64: _backward_

dcicutils/s3_utils.py
```diff
@@ -394,6 +354,125 @@ def unzip_s3_to_s3(self, zipped_s3key, dest_dir, acl=None, store_results=True):
...
+        # we found a match, so fetch that file as config
+        get_response = s3_client.get_object(Bucket=global_bucket, Key=env)
+        env_config = json.loads(get_response['Body'].read())
+        return env_config
+
+    @staticmethod
+    def fetch_health_page_json(url, use_urllib):
+        # TODO: Does anyhone know what problem this use_urllib thing was solving?  Was it solving some bug
+        #       in Python 2 that no longer matters? These two solutions should be equivalent. Is there a way
+        #       to simplify this? -kmp 22-Aug-2021
```
willronchetti, line 448: `requests` does not work (give it a try). No idea why.

test/conftest.py
```diff
@@ -13,13 +14,14 @@
...
+def _discover_es_health_from_boto3_eb_metadata(envname):
     try:
         eb_client = boto3.client('elasticbeanstalk', region_name=REGION)
         # Calling describe_beanstalk_environments is pretty much the same as doing eb_client.describe_environments(...)
         # except it's robust against AWS throttling us for calling it too often.
-        envs = describe_beanstalk_environments(eb_client, EnvironmentNames=[envname])['Environments']
-        for env in envs:
+        envs_from_eb = describe_beanstalk_environments(eb_client, EnvironmentNames=[envname])['Environments']
+        for env in envs_from_eb:
+            print(f"Checking {env.get('EnvironmentName')} for {envname}...")
```
willronchetti, line 24: `PRINT`?

test/conftest.py
```diff
@@ -30,41 +32,50 @@ def _discover_es_health(envname):
...
+    try:
 
-    # This used to be:
-    # INTEGRATED_ES = 'https://search-fourfront-mastertest-wusehbixktyxtbagz5wzefffp4.us-east-1.es.amazonaws.com'
-    # but it changes too much, so now we discover it from the 'elasticsearch' and 'namespace' parts of health page...
-    INTEGRATED_ES, INTEGRATED_ES_NAMESPACE = _discover_es_info(INTEGRATED_ENV)
+        discovered_health_json_from_eb = _discover_es_health_from_boto3_eb_metadata(integrated_envname)
+        assert discovered_health_json_from_eb, f"No health page for {integrated_envname} was discovered."
+        print(f"in _discover_es_url_from_boto3_eb_metadata,"
+              f"discovered_health_json_from_eb={json.dumps(discovered_health_json_from_eb, indent=2)}")
```
willronchetti, line 42: `PRINT`?

willronchetti (approved):
Approving but should be tested in fourfront and CGAP (both legacy and ECS)

dcicutils/s3_utils.py
```diff
@@ -394,6 +354,125 @@ def unzip_s3_to_s3(self, zipped_s3key, dest_dir, acl=None, store_results=True):
...
+
+    LEGACY_ES_URL_KEY = 'es'
+    ES_URL_KEY = 'es_url'
+
+    LEGACY_ENV_NAME_KEY = 'ff_env'
+    ENV_NAME_KEY = 'env_name'
+
+    def __init__(self, env_name=None, env_description=None, s3=None):
+        self.s3 = s3 or boto3.client('s3')
+        if env_description and env_description:
```
aschroed, line 370: Typo?  One should be env_name maybe?

dcicutils/s3_utils.py
netsettler: Thanks. Fixed.

dcicutils/s3_utils.py
aschroed: I don't think it's that requests doesn't work - just that there is no json representation returned for the health page (same sort of thing for indexing_status and perhaps some other endpoints?

And in that case maybe the whole requests clause should just be removed and always use urllib?

test/conftest.py
netsettler: Fixed.

test/conftest.py
netsettler: Fixed.

dcicutils/s3_utils.py
netsettler: Yep. Good catch. Fixed now.

dcicutils/s3_utils.py
netsettler: Well, there _is_ a JSON representation returned when `?format=json` is given, or there's supposed to be. But I guess you're right if it fails to do that, that could be the problem. It's possible there is someplace in which tool is defaulting or failing to default a `Content-Type` that is confusing things. Anyway, I removed the version that does `requests.get` and left the one that does `urllib.request.urlopen`.

aschroed (approved):

PR #157, opened 2021-09-15
Allow legacy short name env arguments to S3Utils (C4-706)

netsettler:
This fixes bug [dcicutils 2.3.0 breaks s3Utils use of env short names (C4-706)](https://hms-dbmi.atlassian.net/browse/C4-706).


willronchetti (approved):
Works for me!

SooLee (approved):

PR #160, opened 2021-09-22
Require at least Python 3.6.1 and finally remove various deprecations

netsettler:
This info is also in the changelog:

The major version bump is to allow removal of some deprecated items and to further constrain the Python version.

Strictly speaking, this is an **INCOMPATIBLE CHANGE**, though we expect little or no impact.

In particular, searches of all `4dn-dcic` and `dbmi-cgap` repositories on GitHub show that three places are impacted:

* the `torb` repository is impacted, and since that repo is not in active use, we're not worried about that. 
* the `4dn-cloud-infra` repo has one change it needs made (called out below) relating to `TIBANNA_OUTPUT_HEALTH_PAGE_KEY`.
* the `cgap-portal` library has several `HEALTH_PAGE_KEY` uses in `root.py` and its test file that need to be changed.

In all three cases, minor code adjustments would fix the problem uses allowing uses of version 3.0 or higher.

Specifics:

* Supports versions of Python starting with 3.6.1 and below 3.8.

* Removes support for previously-deprecated function name `whodaman`, which only `torb` was still using. `compute_ff_prd_env` can be used as a direct replacement.

* Removes support for previously-deprecated variable `MAGIC_CNAME` which no one was using any more.

* Removes support for previously-deprecated variable `GOLDEN_DB` which only `torb` was still using. `_FF_GOLDEN_DB` could be used as a direct replacement in an emergency, but only for legacy environments. This is not a good solution for orchestrated environments (C4-689).

* The variables `FF_MAGIC_CNAME`, `CGAP_MAGIC_CNAME`, `FF_GOLDEN_DB`, and `CGAP_GOLDEN_DB`, which had no uses outside of `dcicutils` itself, now have underscores ahead of their names to emphasize that they are internal to `dcicutils` only. `_FF_MAGIC_CNAME`, `_CGAP_MAGIC_CNAME`, `_FF_GOLDEN_DB`, and `_CGAP_GOLDEN_DB`, respectively, could be used as a direct replacement in an emergency, but only for legacy environments. This is not a good solution for orchestrated environments (C4-689).

* The function name `use_input` has been renamed `prompt_for_input` and the preferred place to import it from is now `misc_utils`, not `beanstalk_utils`. (This is just a synonym for the poorly named Python function `input`.)

* The previously-deprecated class name `deployment_utils.Deployer` has been removed. `IniFileManager` can be used as a direct replacement.

* The previously-deprecated function name `guess_mirror_env` has been removed. `get_standard_mirror_env` can be used as a direct replacement.

* The deprecated function name `hms_now` and the deprecated variable name `HMS_TZ` have been removed. `ref_now` and `REF_TZ`, respectively, can be used as direct replacements.

* These previously-deprecated `s3_utils.s3Utils` class variables have been removed:

  * `s3Utils.SYS_BUCKET_HEALTH_PAGE_KEY` replaced by `HealthPageKey.SYSTEM_BUCKET`
  * `s3Utils.OUTFILE_BUCKET_HEALTH_PAGE_KEY` replaced by `HealthPageKey.PROCESSED_FILE_BUCKET`
  * `s3Utils.RAW_BUCKET_HEALTH_PAGE_KEY` replaced by `HealthPageKey.FILE_UPLOAD_BUCKET`
  * `s3Utils.BLOB_BUCKET_HEALTH_PAGE_KEY` replaced by `HealthPageKey.BLOB_BUCKET`
  * `s3Utils.METADATA_BUCKET_HEALTH_PAGE_KEY` replaced by `HealthPageKey.METADATA_BUNDLES_BUCKET`
  * `s3Utils.TIBANNA_OUTPUT_BUCKET_HEALTH_PAGE_KEY` replaced by `HealthPageKey.TIBANNA_OUTPUT_BUCKET`

  Among `4dn-dcic` repos, there was only one active use of any of these, `TIBANNA_OUTPUT_BUCKET_HEALTH_PAGE_KEY`, in `src/commands/setup_tibanna.py` in `4dn-cloud-infra`. It will need to be rewritten.

  Among `dbmi-bgm` repos, all are mentioned only in `src/encoded/root.py` and `src/encoded/tests/test_root.py`, but rewrites to use `HealthPageKey` attributes will be needed there as well.


willronchetti (approved):
These changes all look reasonable. We should confirm Python 3.5 deprecation with others in the group (I don't think it will be an issue).

netsettler:
> These changes all look reasonable. We should confirm Python 3.5 deprecation with others in the group (I don't think it will be an issue).

I previously asked @sbreiff  and @aschroed long ago what they thought.  Andy was the one who said make it 3.6.1 and not just 3.6. Snovault will need adjusting to require 3.6.1 instead of 3.6, but since both Fourfront and CGAP both require 3.6.1, that's no big deal.

@B3rse and @phil-grayson, do you have any reason to suppose 3.6.1 is an unreasonable requirement for `dcicutils`?

Unless  you know of a Tibanna issue, I think we're all set.  And even then, we need to move to very 3.7 soon, so we need to just start forcing the issue of being closer to up-to-date. If this causes problems somewhere, whatever can't tolerate 3.6.1 is what needs attention.

PR #179, opened 2022-01-09
Small changes 2022-01-09

netsettler:
* In `misc_utils`:

  * Add `string_list`.
  * Add `string_md5`.


netsettler:
I accidentally pushed the tag already, so this PR would like to go in exactly as-is if there is not a big problem. If there is a problem of any kind, I'll need to bump the patch version before merging.

willronchetti (approved):

PR #180, opened 2022-01-11
Expand ECS utils

willronchetti:
- Adds some simple functions for task API operations (for triggering deployment action via foursight)

### DIFF
--- a/dcicutils/ecs_utils.py
+++ b/dcicutils/ecs_utils.py
@@ -43,3 +43,25 @@ class ECSUtils:
         for service_name in self.list_ecs_services(cluster_name=cluster_name):
             self.update_ecs_service(cluster_name=cluster_name, service_name=service_name)
         return True
+
+    def list_ecs_tasks(self):
+        """ Lists all available ECS task definitions. """
+        return self.client.list_task_definitions().get('taskDefinitionArns', [])
+
+    def run_ecs_task(self, *, cluster_name, task_name, subnet, security_group):
+        """ Runs the given task name on the given cluster. """
+        return self.client.run_task(
+            cluster=cluster_name,
+            count=1,
+            taskDefinition=task_name,
+            networkConfiguration={
+                'awsvpcConfiguration': {
+                    'subnets': [
+                        subnet
+                    ],
+                    'securityGroups': [
+                        security_group
+                    ]
+                }
+            }
+        )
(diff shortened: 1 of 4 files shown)


test/test_ecs_utils.py
```diff
@@ -26,6 +26,33 @@ def mock_list_ecs_services(*, cluster):
     }
 
 
+def mock_list_ecs_tasks():
```
netsettler, line 29: This seems fine to have, though I'd rather it be part of a MockECSClient class that the MockBoto3Client could issue up. There is a mock GUID dispenser, and a fresh date could be generated on each call, and then this would be suitable for other uses. But I don't see a use of it getting introduced. Is that still coming? Neither `list_ecs_tasks` nor `run_ecs_task` seems to have unit tests.  I could do that generalization later, though. The functions seem simple enough and this is probably fine for now.

dcicutils/ecs_utils.py
```diff
@@ -43,3 +43,25 @@ def update_all_services(self, *, cluster_name):
         for service_name in self.list_ecs_services(cluster_name=cluster_name):
             self.update_ecs_service(cluster_name=cluster_name, service_name=service_name)
         return True
+
+    def list_ecs_tasks(self):
+        """ Lists all available ECS task definitions. """
+        return self.client.list_task_definitions().get('taskDefinitionArns', [])
+
+    def run_ecs_task(self, *, cluster_name, task_name, subnet, security_group):
+        """ Runs the given task name on the given cluster. """
```
netsettler, line 52: Expand doc string to say what the requirements are for subnet and security group, maybe. Questions I had: Does it create an entity (an ecs container instance?) and give it these attributes on a one-time basis? Is the cluster created or does it have to exist already? I think from the boto3 doc that it exists already? And the network configuration refers to existing resources but you can't run get a network interface without confirming that it's OK to use these existing things? I'm not expecting a book-length doc string, and maybe even the details of how it does this is not needed. But saying these are existing resources, not new ones, is probably good. And also mentioning that this only starts the task but does not wait for a result is helpful. In my view, the AWS operation is badly named but since you're using their name, I won't quibble. They seem to have two different ways to start a task and picked this because 'start' was already taken.

netsettler (approved):
My comments are not blocking.

PR #182, opened 2022-01-20
Python 3.8 compatibility (C4-761)

netsettler:
Python 3.8 compatibility ([C4-761](https://hms-dbmi.atlassian.net/browse/C4-761)):

 * So far, just changing the declaration to let it happen seems to work.


willronchetti (approved):
Works for me with JH in Python 3.8

PR #184, opened 2022-02-04
Python 3.9 compatibility for dcicutils (C4-774)

netsettler:
Python 3.9 compatibility ([C4-774](https://hms-dbmi.atlassian.net/browse/C4-774)):

This pretty much just expects it to work. We'll see how that goes.


willronchetti (approved):

netsettler:
@sbreiff says that SubmitCGAP works in 3.9. It uses this support.

netsettler:
I fixed a unit test that had broken because it depended on `fourfront-cgapdev`, which was recently decommissioned.

PR #186, opened 2022-03-02
New creds_utils migrated from SubmitCGAP (C4-778)

netsettler:
* Adds support for `creds_utils`.

  * Class `KeyManager`, with methods:

    * `KeyManager.get_keydict_for_env(self, env)`

    * `KeyManager.get_keydict_for_server(self, server)`

    * `KeyManager.get_keydicts(self)`

    * `KeyManager.get_keypair_for_env(self, env)`

    * `KeyManager.get_keypair_for_server(self, server)`

    * `KeyManager.keydict_to_keypair(auth_dict)`

    * `KeyManager.keypair_to_keydict(auth_tuple, *, server)`

  * Class `FourfrontKeyManager`

  * Class `CGAPKeyManager`

Opportunistic:

* Adds a `clear-poetry-cache` target in `Makefile`

### DIFF
--- a/dcicutils/creds_utils.py
+++ b/dcicutils/creds_utils.py
@@ -1,0 +1,191 @@
+# import contextlib
+# import functools
+import io
+import json
+import os
+
+# from dcicutils.misc_utils import local_attrs
+from dcicutils.exceptions import AppEnvKeyMissing, AppServerKeyMissing
+
+
+# LOCAL_SERVER = "http://localhost:8000"
+# LOCAL_PSEUDOENV = 'fourfront-cgaplocal'
+#
+# PRODUCTION_SERVER = 'https://cgap.hms.harvard.edu'
+# PRODUCTION_ENV = 'fourfront-cgap'
+#
+# DEFAULT_ENV_VAR = 'SUBMITCGAP_ENV'
+#
+# keys_file_var='CGAP_KEYS_FILE'
+
+
+class KeyManager:
+
+    APP_NAME = None
+    APP_TOKEN = None  # Set this to a string to override APP_NAME.lower() as the app token
+
+    DEFAULT_ENV = None
+    DEFAULT_ENV_VAR = None
+
+    KEYS_FILE = None
+    KEYS_FILE_VAR = None
+
+    # Instance Methods
+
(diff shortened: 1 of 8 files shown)


dcicutils/creds_utils.py
```diff
@@ -0,0 +1,272 @@
...
+        key_manager_class = _KEY_MANAGERS.get(name)
+        if not key_manager_class:
+            raise ValueError(f"There is no registered KeyManager class named {name!r}.")
+        return key_manager_class(**kwargs)
+
+    # @contextlib.contextmanager
+    # def alternate_keys_file_from_environ(self):
+    #     filename = os.environ.get(self.KEYS_FILE_VAR) or None  # Treats empty string as undefined
+    #     with self.alternate_keys_file(filename=filename):
+    #         yield
```
willronchetti, line 129: Several functions are commented out, it isn't necessarily clear which are truly intended to be.

dcicutils/creds_utils.py
```diff
@@ -0,0 +1,272 @@
...
+        old_filename = cls.KEYS_FILE
+        try:
+            cls.KEYS_FILE = filename or cls._default_keys_file(for_testing=True)
+            yield
+        finally:
+            cls.KEYS_FILE = old_filename
+
+    @classmethod
+    def register(cls, *, name):
+        def _register_class(key_manager_class):
```
willronchetti, line 74: I'd add docstrings especially to functions like these where the syntax/intended structure/use are not immediately obvious (to all) by reading it

willronchetti (approved):
This all looks great, very useful refactor that will hopefully centralize our use of this sort of utility

dcicutils/creds_utils.py
netsettler: Ah, yes. SubmitCGAP had these as functions but I think they were not needed, so I simplified the paradigm. Basically, some things want to be class variables but need inits, so what I did to initialize them functionally was to use a decorator that runs after the class definition and knows how to do the initializations. I think it may be that in later Python versions you can write an init class kind of thing, or maybe I've just seen it as something users can write using metaclasses, but I think the decorator solution is adequate and in some ways easier to think about.

PR #189, opened 2022-04-11
New mappings and noramalizer support for diff_utils

netsettler:
* Add support for `include_mappings=True` and `normalizer=<fn>` in `DiffManager.diffs`.
* Backfill `CHANGELOG` documentation for the change in dcicutils 3.11.1

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,25 @@ Change Log
 ----------
 
 
+3.12.0
+======
+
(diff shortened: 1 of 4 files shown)


netsettler:
There are some test failures, but they fail on master, too, so they are likely due to changes in mastertest data that require a new PR to fix, and not due to something changed in this PR. Relevant tests for this PR are passing.

PR #190, opened 2022-04-24
A few changes in support of recent containers work

netsettler:
* In `deployment_utils`:

  * In `CreateMappingOnDeployManager`:

    * Add `get_deploy_config` with slightly different arguments than `get_deployment_config`, so CGAP and FF can be more easily compared.

    * Change `--strict` and `--skip` to not take an argument on the command line, and to default to False.

      NOTE: After some discussion, this was decided to be treated as a bug fix, not as an incompatible change, so the semantic version will not require a major version bump.

  * When testing, test with switch arguments, too.

* In `env_utils`:

  * Add `FF_ENV_PRODUCTION_BLUE_NEW` (value `'fourfront-production-blue'`) and `FF_ENV_PRODUCTION_GREEN_NEW` (value `'fourfront-production-green'`) and teach `is_stg_or_prd_env` and `get_standard_mirror_env` about them as alternative stg/prd hosts.

  * Add `is_beanstalk_env` to detect traditional/legacy beanstalk names during transition.

* In `qa_utils`:

  * Add `MockedCommandArgs`.


dcicutils/env_utils.py
```diff
@@ -56,6 +59,25 @@ def _orchestrated_app_case(orchestrated_app: OrchestratedApp, if_cgap, if_fourfr
 # CGAP_ENV_WEBPROD2_NEW is meaningless here. See CGAP_ENV_STAGING_NEW.
 CGAP_ENV_WOLF_NEW = 'cgap-wolf'  # Maybe not used
 
+_ALL_BEANSTALK_NAMES = [
+    # You can peek at contents of this variable to see how is_beanstalk_env will work,
+    # but DO NOT reference this directly from code outside of th is module. -kmp 18-Apr-2022
```
willronchetti, line 64: extra space in comment ("... of th is module.")

dcicutils/env_utils.py
```diff
@@ -56,6 +59,25 @@ def _orchestrated_app_case(orchestrated_app: OrchestratedApp, if_cgap, if_fourfr
...
+    # Fourfront beansstalks
+    'fourfront-webprod',      # FF_ENV_WEBPROD (superseded by fourfront-green)
+    'fourfront-webprod2',     # FF_ENV_WEBPROD2 (superseded by fourfront-blue)
+    'fourfront-green',        # FF_ENV_PRODUCTION_GREEN
+    'fourfront-blue',         # FF_ENV_PRODUCTION_BLUE
+    'fourfront-mastertest',   # FF_ENV_MASTERTEST
+    'fourfront-webdev',       # FF_ENV_WEBDEV
+    'fourfront-hotseat',      # FF_ENV_HOTSEAT
+    # CGAP beanstalks
+    'fourfront-cgap',         # CGAP_ENV_WEBPROD (decomissionsed, replaced by AWS orchestration)
```
willronchetti, line 75: Decommissioned?

willronchetti (approved):

dcicutils/env_utils.py
netsettler: fixed

dcicutils/env_utils.py
netsettler: Ugh. Several of these. Fixed. Thanks.

netsettler:
I've pushed up a set of tests that pass locally on my machine. They seem to pass on GA as well, so now we have something stable to test against as we decommission the fourfront beanstalks over the next week. Please report spontaneous failures that emerge without source changes beyond this PR, as they will probably be due to config changes.

PR #193, opened 2022-05-16
Kmp alt2 env utils unit testing recorded

netsettler:
Changes to allow converting integration tests to unit test by recording them.

In order to create a recording, one must use 
```
RECORDING_ENABLED=TRUE pytest -vv -m recordable
```

There are three interesting cases:

* Unit test is run.  This uses the most recent recording.
* Integration test is run with `RECORDING_ENABLED=TRUE`. This does an integration test, recording it for future use.
* Integration test is run with `RECORDING_ENABLED` left to default. This runs the integration test with no recording.

NOTE: I tried using `ControlledTime` to speed this up, but that didn't work. I suspect there is some complexity to mocking that I am missing. So this is slower than I wish, but at least not dependent on a remote host for the recorded values.


willronchetti (approved):
Cool utility - could be useful for other mechanisms, like `VirtualApp`

PR #194, opened 2022-05-23
Upgrades related to deployment_utils.py

netsettler:
This is a branch-to-branch PR that adds some deployment_utils changes to `kmp_alt2_env_utils_unit_testing`.

This allows new GAC settings:

* `GLOBAL_ENV_BUCKET` (which may already be set but won't get picked up by deployment)
* `ENCODED_ECOSYSTEM` (which is new)
* `ENCODED_ENV_NAME` (which can replace ENCOED_BS_ENV, or if that causes problems can be used along side it with a consistent value)

In .ini files, this will create these corresponding environment variables:

* `ENV_BUCKET`
* `ENV_ECOSYSTEM`
* `ENV_NAME` (though `BS_ENV` exists as well for compatibility)

and should be used to install these values in the registry:

* `env.bucket`
* `env.ecosystem`
* `env.name` (already exists)

I also use `env.name.bs` in testing just as a place to put `BS_ENV` so that I can test that `ENV_NAME` and `BS_NAME` are getting the same value generated, but nothing expects that value downstream and the marker should not be used as a model for anything to put in the portal `.ini` files.

This other stuff (`.env.bucket`, `.env.ecosystem`, and `env.name`) will work in conjunction with a change that's soon pending to `cgap-portal` to find these values and present them in the health page.  I'm not sure anything else cares about them, but it will be useful to see them documented so that people trying to configure the system will know what values to put. 

For possible future work (not in this PR):

* `4dn-cloud-infra` might be made to use this as a cross-check at some point.
* `foursight` can use it as something to bootstrap from.
* We might want to extend `s3_utils.s3Utils` to contain or check this information.


test/ini_files/blue.ini
```diff
@@ -9,7 +9,10 @@ system_bucket = elasticbeanstalk-fourfront-webprod-system
 accession_factory = encoded.server_defaults.enc_accession
 elasticsearch.server = search-fourfront-blue-xkkzdrxkrunz35shbemkgrmhku.us-east-1.es.amazonaws.com:80
 snovault.app_version = ask-pip
+env.bucket = foursight-envs
+env.ecosystem = main
 env.name = fourfront-blue
+env.name.bs = fourfront-blue
```
willronchetti, line 15: In theory all these files can be removed except the any.ini versions (which should be aligned to the versions in cgap-portal)

test/ini_files/cgap_alpha.ini
```diff
@@ -20,7 +20,10 @@ s3_encrypt_key_id = ${S3_ENCRYPT_KEY_ID}
 accession_factory = encoded.server_defaults.enc_accession
 elasticsearch.server = search-fourfront-cgap-ewf7r7u2nq3xkgyozdhns4bkni.us-east-1.es.amazonaws.com:80
 snovault.app_version = ask-pip
+env.bucket = foursight-envs
```
willronchetti, line 23: Not a sensible fallback for CGAP I'd say... Not that a good one exists. If this is blank will it throw an error? This bucket should always be available when the ECS comes up so I think it's okay to require. Same comment in other cgap ini files.

willronchetti (approved):

test/ini_files/cgap_alpha.ini
netsettler: Well, this is just a sample file, so this isn't the place where the fallback is computed. I'll make an adjustment to the defaulting, though.

test/ini_files/blue.ini
netsettler: Yeah, I'll remove those later when I'm sure things are in a working state and I have time to consider a simpler replacement

PR #195, opened 2022-06-03
Fix empty containers problem in diff_utils (C4-838)

netsettler:
In most of diff_utils, it treats {} and {"foo": {}} as essentially the same. For patch_utils, this is not good, but for now I'll leave the other stuff stable.

The tests don't run on this, so I added a function interim_run_all() that runs all the diff_utils tests and they do pass on master.

### DIFF
--- a/dcicutils/diff_utils.py
+++ b/dcicutils/diff_utils.py
@@ -59,7 +59,7 @@ class DiffManager:
 
     CONSTANTS_MAP = {None: 'null', True: 'true', False: 'false'}
 
-    def unroll(self, item, normalizer=None, _omit_subscripts=False):
+    def unroll(self, item, normalizer=None, _omit_subscripts=False, _omit_empty_containers=True):
         """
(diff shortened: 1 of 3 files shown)


dcicutils/diff_utils.py
```diff
@@ -140,7 +145,7 @@ def patch_diffs(self, item):
         """
         Returns a list of keys for would-be-affected properties if item were a patch request for a piece of JSON.
         """
-        result = self.diffs({}, item, _omit_subscripts=True)
+        result = self.diffs({}, item, _omit_subscripts=True, _omit_empty_containers=False)
```
willronchetti, line 148: I'm fine with defaulting to the desired behavior, but this should probably be an argument to this function as well.

willronchetti (approved):

dcicutils/diff_utils.py
netsettler: I'll take that up another time. I'm going to merge this as-is. The interim test function says tests pass.

PR #197, opened 2022-06-17
env_xxx refactor (compatible)

netsettler:
This PR only changes the structure of some files (in a way that makes other diffs hard). It does not change functionality that exists already. 

Basically, this will divide EnvManager into a lower level `EnvBase` class (in `env_basic.py`) and an `EnvManager` class (in `env_manager.py`) that inherits from `EnvBase`. What makes this a compatible change is that, having emptied out the previous contents of `env_base.py` into these two other files, I then re-import `EnvManager` back into `env_base.py` so that anything doing `from dcicutils.env_base import EnvManager` can still find it.

Aside to @dmichaels-harvard, who might not be familiar with this technique, the `exported` function in this file is entirely a no-op that I invented. The file contents that are now:
```
from dcicutils.misc_utils import exported
from .env_manager import EnvManager

exported(EnvManager)  # It used to live here
```
would work just as well if `env_base.py` only contained:
```
from .env_manager import EnvManager
```
but it would be harder to see what was going on. PyCharm and other linters would think the import was unused, but that's because they don't scan other files for
```
from dcicutils.env_base import EnvManager
```

I would _prefer_ to make the incompatible change that would break this, and I'll make a layered PR that offers that change instead so you can choose that, but that will create cascade work in other repos in order to take this change.


dcicutils/ff_mocks.py
```diff
@@ -135,37 +135,38 @@ def mocked_s3utils(environments=None, require_sse=False, other_access_key_names=
         with mock.patch.object(ff_utils, "boto3", mock_boto3):
             with mock.patch.object(beanstalk_utils, "boto3", mock_boto3):
                 with mock.patch.object(env_utils, "boto3", mock_boto3):
```
netsettler, line 137: The diff here looks complicated but it's just latching onto the wrong thing. Really all that's changed is
```
with mock.patch.object(env_base, "boto3", mock_boto3):
```
became
```
with mock.patch.object(env_basic, "boto3", mock_boto3):
    with mock.patch.object(env_manager, "boto3", mock_boto3):
```
and everything that followed is indented four extra spaces.

netsettler:
This PR was superseded by [utils PR #198](https://github.com/4dn-dcic/utils/pull/198).

PR #198, opened 2022-06-17
env_xxx refactor (incompatible)

netsettler:
This PR changes the structure of some files (in a way that makes other diffs hard).

This PR is slightly incompatible with v3.14.0.2b38, and will require that `EnvManager` be imported from `dcicutils.env_manager` rather than `dcicutils.env_base`. However, `dcicutils.env_base` is an artifact of this beta series and is not present on master. Moreover, we are planning to bump the `dcicutils` major version already, so incompatible changes are really only an issue for repos conspiring in the beta.

Other than forcing imports to change, there is not a functionality change caused by the rest of this PR. Ask me if you have questions.

Either this PR or [env_xxx (compatible)](https://github.com/4dn-dcic/utils/pull/197) should be selected so that other changes can be considered against one of those.


willronchetti (approved):
This refactor looks reasonable to me

PR #200, opened 2022-06-19
Changes in service of foursight-core (incompatible)

netsettler:
This PR is built on [PR #198: env_xxx refactor (incompatible)](https://github.com/4dn-dcic/utils/pull/198), which factors `EnvManager` out into separate file `dcicutils/env_manager.py`, and leaves only a more basic bit of functionality as `EnvBase` in `dcicutils.env_base.py`.

The big changes are:
* In `dcicutils/env_manager.py` (lines 166/155): Change `verify_and_get_env_config` to call `full-env-name` on its `env` argument so that all three names (public, short and long) can be given without having to have 3 different declarations in the global env bucket. NOTE: This change is the one that forced the env refactor. 
* In `dcicutils/env_utils.py` (lines 536/569): There is a configuration change to add `"default_workflow_env"` as a configuration option. (The default was erroneously going to the `"webprod_pseudo_env"`.)
* In `dcicutils/env_utils.py` (lines 895/928): There is an additional argument to `infer_foursight_from_env` and some additional defaulting of the result.

Lesser changes, still of note:
* In `dcicutils/env_utils.py`, some new methods on `EnvUtils` support local uses of `EnvUtils` in various modalities.
* In `dcicutils/ff_mocks.py`, `mocked_s3utils` now automatically mocks up a minimal ecosystem file.
* In `dcicutils/ff_mocks.py`, new class `IntegratedFixture` supports the `integrated_ff` fixture in a way that more reliably assures initialization of certain quantities happens at file load time. Along the way, this also fixes [Test credentials appear in stack trace on GA (C4-847)](https://hms-dbmi.atlassian.net/browse/C4-847).
* In `test/conftest.py` and `test/conftest_settings.py`, use new `IntegratedFixture` class.
* In `test/helpers.py`, some additional context managers.
* In `test/test_deployment_utils.py`, in addition to fixing some tests, I got rid of some old useless ones and updated them to do something more relevant instead.
* In `test/test_s3_utils.py`, repaired an expectation that `staging` would use `http` so it now expects `https`.
* Added a bunch of context decorators in various files to make tests not be so sensitive to orrder.


willronchetti (approved):
These changes all look reasonable as well

PR #201, opened 2022-06-19
Mocking print to file

netsettler:
This is just a random piece of functionality I half-wrote while talking to David about the print mocking we already had. It seemed potentially useful, so I wanted to go ahead and install it for possible future use.

Because unit testing is generally broken, more specific testing can be done by:
```
>>> from test.test_misc_utils import interim_run_all
>>> interim_run_all()
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,17 @@ Change Log
 ----------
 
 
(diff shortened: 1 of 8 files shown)


willronchetti (approved):
Looks useful!

PR #203, opened 2022-07-06
Synthetic foursight decls

netsettler:
This change lets the portal and es urls needed by foursight get inferred from the ecosystem declaration so the environment declaration really only needs the ecosystem. The changes are small to the main system, a bit more complicated to tests. It needs careful review.

In the process, `infer_foursight_url_from_env` and `infer_foursight_from_env` from `env_utils` were made to require keyword arguments since I found a case or two where one arg was passed and it was assumed it was the env name but it was being taken as a request instead. Some adjustments in `foursight-core` or `foursight-cgap` might be needed, as this is an incompatible change, but they'll be small.

### DIFF
--- a/test/test_deployment_utils.py
+++ b/test/test_deployment_utils.py
@@ -574,7 +574,8 @@ def test_deployment_utils_transitional_equivalence():
         # TODO: Once this mechanism is in place, the files cgap.ini, cgapdev.ini, cgaptest.ini, and cgapwolf.ini
         #       can either be removed (and these transitional tests removed) or transitioned to be test data.
 
-        def tester(ref_ini, env_name, data_set, es_server, *, any_ini=None, es_namespace=None, line_checker=None,
+        def tester(ref_ini, env_name, data_set, es_server, *,
+                   higlass_server=None, any_ini=None, es_namespace=None, line_checker=None,
                    use_ini_file_manager_kind=None, **others):
             """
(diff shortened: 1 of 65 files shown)


netsettler:
This is a branch-to-branch PR so doesn't auto-test. I launched tests manually and [they pass](https://github.com/4dn-dcic/utils/actions/runs/2624825705) (now that I fixed some things the first launch found).

test/test_deployment_utils.py
```diff
@@ -1452,6 +1457,71 @@ def test_get_deployment_config_ff_hotseat_orchestrated():
...
+        finally:  # show even if we err
+            print(f"{info} env={env.ljust(20)} wipe_es={'Y' if wipe_es else 'N'}"
+                  f" full={full_env_name(env).ljust(20)} kind={kind}")
+        return config
+
+    print()  # Start on a fresh line
+    for wipe_es in [False, True]:
+        print(f"wipe_es={wipe_es}")
+        for kind_and_envs in [['production', ['fourfront-prd', 'prd', 'data']],
+                              ['staging', ['fourfront-stg', 'stg', 'staging']],
```
willronchetti, line 1496: Seems like `fourfront-production-green` and `fourfront-production-blue` should be here?

willronchetti (approved):
Approving this for merge into main EnvUtils PR

PR #205, opened 2022-07-13
New MockBoto3Ec2 class in qa_utils geared toward security group rule…

dmichaels-harvard:
…s unit testing; and associated tests; and also tests for MockBoto3Session related to credentials.

### DIFF
--- a/test/test_qa_utils_mock_boto3_session.py
+++ b/test/test_qa_utils_mock_boto3_session.py
@@ -7,13 +7,13 @@ import tempfile
 import uuid
 
 
-def test_mock_boto3_session():
+def test_mock_boto3_session() -> None:
     mock_boto3 = MockBoto3()
     assert isinstance(mock_boto3.client('session'), MockBoto3Session)
     assert isinstance(mock_boto3.client('session', region_name='us-east-1'), MockBoto3Session)
 
 
-def test_mock_boto3_session_get_credentials_via_explicit():
+def test_mock_boto3_session_get_credentials_via_explicit() -> None:
     mock_boto3 = MockBoto3()
     mocked_session = mock_boto3.session.Session()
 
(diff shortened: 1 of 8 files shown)


test/test_qa_utils_mock_boto3_session.py
```diff
@@ -0,0 +1,102 @@
...
+    aws_access_key_id = str(uuid.uuid4())
+    aws_secret_access_key = str(uuid.uuid4())
+    aws_region = str(uuid.uuid4())
+    mocked_session.unset_environ_credentials_for_testing()
+    os.environ["AWS_ACCESS_KEY_ID"] = "do-not-find-this-access-key-id"
+    os.environ["AWS_SECRET_ACCESS_KEY"] = "do-not-find-this-secret-access-key"
+    mocked_session.put_credentials_for_testing(aws_access_key_id=aws_access_key_id,
+                                               aws_secret_access_key=aws_secret_access_key,
+                                               region_name=aws_region,
+                                               aws_credentials_dir=None)
```
netsettler, line 30: Whether implicitly by an unset method or explicitly by assigning things in os.environ assignments, if this is going to change environment variables, you need to do this kind of thing inside a context manager that will put the variables back at the end, or else running this test will leave the variables set weirdly for other tests. Keep in mind, too, that there could be an error at any point in the test and you won't execute the rest of the test, so if you set them at the end, that's insufficiently. It has to be done by a try/finally.  So probably the unset credentials thing should be reworked as a context manager, perhaps implemented through the `misc_utils.override_environ` context manager, which itself takes care of putting things back when it's done.

dmichaels-harvard: Oh, right, duh, I should have known that, thanks, will take care of it.

willronchetti (approved):
Nice! Should be very useful

PR #208, opened 2022-07-18
Enhancements to identity management tools

netsettler:
This refactors the way some of the secrets manager code works to add some extra functionality and abstraction.

### DIFF
--- a/dcicutils/secrets_utils.py
+++ b/dcicutils/secrets_utils.py
@@ -8,6 +8,7 @@ import structlog
 from botocore.exceptions import ClientError
 from typing import Optional
 from .common import REGION as COMMON_REGION
+from .lang_utils import a_or_an
 from .misc_utils import PRINT, full_class_name, override_environ
 
 
@@ -76,9 +77,18 @@ def get_identity_secrets(identity_kind=GLOBAL_APPLICATION_CONFIGURATION) -> dict
 assume_identity = get_identity_secrets  # assume_identity is deprecated. Please rewrite
 
 
-def apply_overrides(*, secrets: dict, overrides: Optional[dict]) -> dict:
-    if overrides:
-        secrets = dict(secrets, **overrides)
+def apply_overrides(*, secrets: dict, rename_keys: Optional[dict] = None,
+                    override_values: Optional[dict] = None) -> dict:
+    if rename_keys:
+        secrets = secrets.copy()
+        for key, new_name in rename_keys.items():
+            if new_name in secrets:
+                raise ValueError(f"Cannot rename {key} to {new_name} because {a_or_an(new_name)} attribute already exists.")
+            if key not in secrets:
+                raise ValueError(f"Cannot rename {key} to {new_name} because the secrets has no {key} attribute.")
+            secrets[new_name] = secrets.pop(key)
+    if override_values:
+        secrets = dict(secrets, **override_values)
     return secrets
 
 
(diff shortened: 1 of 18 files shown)


dcicutils/secrets_utils.py
```diff
@@ -53,12 +58,73 @@ def assume_identity():
...
+                yield
+                return  # Nothing to do after that
+
+        elif require_identity:
+            raise RuntimeError(f"An identity was not defined in environment variable {identity_kind}.")
+
+    yield
+
+
+def apply_identity(identity_kind=GLOBAL_APPLICATION_CONFIGURATION):
```
willronchetti, line 113: Definitely very useful

dcicutils/secrets_utils.py
```diff
@@ -20,14 +21,18 @@
 GLOBAL_APPLICATION_CONFIGURATION = 'IDENTITY'
 
 
-def assume_identity():
+def get_identity_name(identity_kind=GLOBAL_APPLICATION_CONFIGURATION):
+    return os.environ.get(identity_kind, 'dev/beanstalk/cgap-dev')
+
+
+def get_identity_secrets(identity_kind=GLOBAL_APPLICATION_CONFIGURATION) -> dict:
```
willronchetti, line 28: I would map this to `assume_identity` to avoid the incompatible change while we track all the usage down.

dcicutils/secrets_utils.py
```diff
@@ -53,12 +58,73 @@ def assume_identity():
...
+
+    Since this is assumed to be an initial startup action, we assume it's not already been done and don't
+    offer the only_if_missing mechanism that the assumed_identity context manager offers.
+
+    """
+    gac = os.environ.get(identity_kind)
+    if gac:
+        secrets = get_identity_secrets(identity_kind=identity_kind)
+        for var, val in secrets.items():
+            os.environ[var] = val
```
willronchetti, line 125: You might consider providing an override mechanism for the keys, where for example if a key is `ENCODED_AUTH0_CLIENT` in the GAC but we want `CLIENT_ID` in foursight, you can specify a dictionary mapping override `ENCODED_AUTH0_CLIENT` --> `CLIENT_ID`

willronchetti (approved):
Approving this with some small comments

dcicutils/secrets_utils.py
netsettler: I think there is an 
```
assume_identity = get_identity_secrets
```
farther down that does exactly this.

dcicutils/secrets_utils.py
netsettler: OK, I think that's not difficult.

dcicutils/secrets_utils.py
```diff
@@ -53,12 +58,73 @@ def assume_identity():
...
+
+def apply_identity(identity_kind=GLOBAL_APPLICATION_CONFIGURATION):
+    """
+    Assumes an identity globally.
+
+    Since this is assumed to be an initial startup action, we assume it's not already been done and don't
+    offer the only_if_missing mechanism that the assumed_identity context manager offers.
+
+    """
+    gac = os.environ.get(identity_kind)
```
dmichaels-harvard, line 121: Couldn't/shouldn't this be a call to: get_identity_name(identity_kind) ... ? (And variable named identity_name rather than gac?)

netsettler:
This passes all tests locally on my machine. And it lints clean. I think I'm now testing all the functions in secrets_utils, and most of their functionality. So I'm going to merge this down into the main beta branch.

PR #209, opened 2022-07-27
Factored out lookup part of AbstractOrchestrationManager.discover_fou…

dmichaels-harvard:
…rsight_check_runner_name into find_lambda_function_names.

### DIFF
--- a/dcicutils/cloudformation_utils.py
+++ b/dcicutils/cloudformation_utils.py
@@ -232,6 +232,37 @@ class AbstractOrchestrationManager:
                     return getattr(summary, attr, default)
         return default
 
+    @classmethod
+    def find_lambda_function_names(cls, name_pattern) -> list:
+        """
+        Returns a list of AWS lambda function names which match the given regular expression,
+        which may be a regular expression string or a compiled one (i.e. of type re.Pattern);
+        returns an empty list if no matches.
+
+        :param name_pattern: Regex pattern to match AWS lambda function name; string or re.Pattern.
+        :return: List of matching AWS lambda function names, or empty list if none.
+        """
+        if not name_pattern:
+            return []
+        elif isinstance(name_pattern, str):
+            name_pattern = re.compile(name_pattern)
+        elif not isinstance(name_pattern, re.Pattern):
+            return []
+        matching_function_names = []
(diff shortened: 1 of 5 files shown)


dcicutils/cloudformation_utils.py
```diff
@@ -232,6 +232,35 @@ def find_stack_resource(self, stack_name_token, resource_logical_id, attr=None,
                     return getattr(summary, attr, default)
         return default
 
+    @classmethod
+    def find_lambda_function_names(cls, name_pattern) -> list:
```
willronchetti, line 236: You might want to test this, I would just do something simple like mock out a response from `list_functions` and verify it gives the functions you were looking for.

willronchetti (approved):

dcicutils/cloudformation_utils.py
netsettler: I think there is support for a lambda client where you specify the names  you want to test against, so that should be easy.

test/test_cloudformation_utils.py
```diff
@@ -489,3 +490,47 @@ def test_abstract_orchestration_manager_discover_foursight_check_runner_name():
                     assert foo_runner == MOCKED_DEV_CHECK_RUNNER
                     assert os.environ.get('CHECK_RUNNER') == foo_runner
                     print(f"CHECK_RUNNER={os.environ.get('CHECK_RUNNER')}")
+
+
+def test_find_lambda_function_names():
+
+    mocked_lambdas = [
+        "c4-foursight-cgap-supertest-stack-CheckRunner-ABC",
+        "c4-foursight-fourfront-production-stac-CheckRunner-DEFGHI",
```
netsettler, line 499: You should insert about 50 generated 'decoy' items in each position before, amidst, and after these so that you're exercising all the nextmarker stuff, forcing things onto different pages.

dmichaels-harvard: Oh yeah good call, and trying it actually running into something, thanks.

dmichaels-harvard: Added a bunch of dummy lambdas to the list (total: 203) to test the nextmarker stuff.

PR #212, opened 2022-08-16
Add search facet cnt

aschroed:
Small change to allow you to grab the facet values and counts from any arbitrary search.
The original get_item_facet_values function returned the facet values and document counts for the specified item_type passed to the function.  
I added a new function get_search_facet_values that does all the same logic using the result of any search and changed the get_item_facet_values function to build a search string that is passed to the new function.

### DIFF
--- a/dcicutils/ff_utils.py
+++ b/dcicutils/ff_utils.py
@@ -469,14 +469,15 @@ def get_item_facets(item_type, key=None, ff_env=None):
     return facets
 
 
-def get_item_facet_values(item_type, key=None, ff_env=None):
+def get_search_facet_values(search_query, key=None, ff_env=None):
     """
-    Gets all facets and returns all possible values for each one with counts
-    ie: dictionary of facets mapping to a dictionary containing all possible values
+    Gets all facets available from a provided search and returns all possible values for
+    each with counts - query should begin 'search/...'
+    ie: returns dictionary of facets mapping to a dictionary containing all possible values
     for that facet mapping to the count for that value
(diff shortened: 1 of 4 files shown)


dcicutils/ff_utils.py
```diff
@@ -485,14 +485,15 @@ def get_item_facets(item_type, key=None, ff_env=None):
     return facets
 
 
-def get_item_facet_values(item_type, key=None, ff_env=None):
+def get_search_facet_values(search_query, key=None, ff_env=None):
```
netsettler, line 488: Should the `search_query` not maybe default more easily?

Untested:
```suggestion
DEFAULT_SEARCH_QUERY_PATTERN = "search/?type=%s"


def get_search_facet_values(search_query=None, item=None, key=None, ff_env=None):
    if (not search_query and not item) or (search_query and item):
        raise ValueError("You must supply exactly one of 'search_query' or 'item')
    elif not search_query:
        search_query = DEFAULT_SEARCH_QUERY_PATTERN % item
```

dcicutils/ff_utils.py
```diff
@@ -502,6 +503,15 @@ def get_item_facet_values(item_type, key=None, ff_env=None):
     return facets
 
 
+def get_item_facet_values(item_type, key=None, ff_env=None):
```
netsettler, line 506: Oh, I see, you're doing it as two separate functions. :)

netsettler (approved):
This is fine as-is. My comments are not blocking.

PR #214, opened 2022-08-18
Small changes in obfuscation_utils and lang_utils

netsettler:
* Add `is_obfuscated`.
* Add `obfuscated=` argument to `obfuscate` and `obfuscate_dict`.
* Add `"from"` and `"between"` to list of prepositions that the pluralizer understands.


dmichaels-harvard (approved):
Look good to me. Thanks.

PR #222, opened 2022-09-29
Improved IntegratedFixture and static check cleanups (C4-932)

netsettler:
* Show less uninteresting traceback info for static tests.
* Small incompatible change to recently released `qa_checkers.confirm_no_uses` to remove the `if_used` argument in favor of a simpler implementation. Since this is testing-only, not something used in production, and since there are believed to be no uses outside the repo, no major version bump.
* Make class initialization of `IntegratedFixture` happen at instance instantiation time. That simplifies the loading actions needed. Those can happen in `conftest.py` rather than in `ff_mocks`, which should allow `dcicutils.ff_mocks` to be imported more easily ([C4-932](https://hms-dbmi.atlassian.net/browse/C4-932)).

### DIFF
--- a/dcicutils/qa_checkers.py
+++ b/dcicutils/qa_checkers.py
@@ -47,7 +47,7 @@ class ChangeLogChecker:
     # on particular versions of pytest, and we don't export a delivery
     # constraint of a particular pytest version. So RuntimeWarning is a
     # safer setting for now. -kmp 14-Jan-2021
-    WARNING_CATEGORY = RuntimeWarning
+    WARNING_CATEGORY: Type[Warning] = RuntimeWarning
 
     @classmethod
     def check_version(cls):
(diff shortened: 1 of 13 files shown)


dcicutils/qa_checkers.py
```diff
@@ -425,7 +432,4 @@ def summarize(problems):
             n += len(matches)
             detail += f"\n In {file}, {summarize(matches)}."
         message = f"{n_of(n, 'problem')} detected:" + detail
-        if if_used == 'warning':
-            warnings.warn(message, category=warning_category, stacklevel=2)
-        else:  # if_used == 'error'
-            raise AssertionError(message)
+        raise AssertionError(message)
```
willronchetti, line 435: Was the `if_used` behavior here removed intentionally? If so docstring should be updated

willronchetti (approved):
Looks good, just one small clarifying question

dcicutils/qa_checkers.py
netsettler: Yeah, it's intentional. It was recently added and no one is using it yet. I found a place higher in the call chain to do this effect more simply. I'll adjust doc string.

PR #223, opened 2022-09-30
Refactored recording tech

netsettler:
This refactors the test recording technology so that it would be easier to make more kinds (e.g., if we want something to use in `cgap-portal`). It may have to still change more later, but @willronchetti and I talked about it and agreed it's better just to freeze it its current form for now so it gets into master and doesn't stuffer bit rot, and worry about that later.

### DIFF
--- a/test/test_ff_mocks.py
+++ b/test/test_ff_mocks.py
@@ -6,12 +6,13 @@ import pytest
 from dcicutils import ff_mocks as ff_mocks_module, ff_utils as ff_utils_module
 from dcicutils.ff_utils import stuff_in_queues
 from dcicutils.ff_mocks import (
-    AbstractIntegratedFixture, AbstractTestRecorder, IntegratedTestRecorder,
+    AbstractIntegratedFixture, AbstractTestRecorder, IntegratedTestRecorder, TestScenarios,
     mocked_authorized_request_verbs, controlled_time_mocking,
 )
 from dcicutils.misc_utils import ignored, local_attrs, PRINT
 from dcicutils.qa_utils import MockResponse, MockFileSystem, ControlledTime, printed_output
 from unittest import mock
+from .helpers import using_fresh_legacy_state_for_testing
 
 
 def test_abstract_integrated_fixture_no_server_fixtures():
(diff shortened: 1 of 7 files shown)


netsettler:
I deployed this to devtest and poked at it a bit. It seems to be operating normally.

test/test_ff_mocks.py
```diff
@@ -1,10 +1,516 @@
...
+
+    assert TestScenarios.mocked_auth_secret('foo') == 'foosecret'
+    assert TestScenarios.mocked_auth_secret('fourfront-foo') == 'foosecret'
+    assert TestScenarios.mocked_auth_secret('fourfront-cgapfoo') == 'cgapfoosecret'
+    assert TestScenarios.mocked_auth_secret('cgap-foo') == 'cgap-foosecret'
+
+    assert TestScenarios.mocked_auth_server('foo') == 'http://foo.4dnucleome.org/'
+    assert TestScenarios.mocked_auth_server('fourfront-foo') == 'http://foo.4dnucleome.org/'
+    assert TestScenarios.mocked_auth_server('fourfront-cgapfoo') == 'http://cgapfoo.4dnucleome.org/'
+    assert TestScenarios.mocked_auth_server('cgap-foo') == 'http://cgap-foo.4dnucleome.org/'
```
willronchetti, line 490: Do we always want a .4dnucleome domain?

willronchetti (approved):

test/test_ff_mocks.py
netsettler: These are test scenarios. It's more a matter of running through the code path. I could just match on a regexp that it returned something like `http://cgap-foo[.].*[.](com|org)/`.

netsettler: I looked at it and think I’ll leave it for now because there are too many cases and the regexp would complicate things. It’s one of those cases where the test is just holding the coverage in check, but also saying “this is what the example returns” not “this is what code has to do”. if someone changes that TestScnaerio it will have to change the test, but that’s not horrible. They don’t change often.

PR #225, opened 2022-10-12
Singleton and classproperty support

netsettler:
The original purpose of thie PR was to add some substrate for use by David in `foursight-core` and probably elsewhere in `4dn-cloud-infra`. We've had a number of needs for both the `classproperty` and `Singleton` support.

In the process, I brought `misc_utils` coverage up to 100% and fixed a few tiny and probably inconsequential bugs that I noticed when adding those tests. 

* In `misc_utils`:

  * New decorator `@classproperty`

  * New decorator `@classproperty_cached`

  * New decorator `@classproperty_cached_each_subclass`

  * New class `Singleton`. Users of `SingletonManager` might prefer this, but we'll continue to support both. (No deprecation for now.)

  * In function `is_valid_absolute_uri`, better handling of argument type errors.

  * For `CachedField`:

    * Added a handler for `__str__` that returns useful information, which can also be used for `__repr__`.

    * Fixed handler for `__repr__` to return a properly executable expression (shared with `__str__`).

  * Improved test coverage by adding tests for some parts of the code that were not previously tested.

* In `qa_utils`:

  * New class `MockId` for mocking the `id` function in a predictable way.

  * Adjust `MOCK_QUEUE_URL_PREFIX` to use a mocked URL that looks more like modern AWS url, where `queue.amazonaws.com` has been replaced by `sqs.us-east-1.amazonaws.com`.

### DIFF
--- a/dcicutils/misc_utils.py
+++ b/dcicutils/misc_utils.py
@@ -1560,6 +1560,110 @@ class SingletonManager:
         return self._singleton_class
 
 
+class classproperty(object):
+    """
+    This decorator is like 'classproperty', but the function is run only on first use, not every time, and then cached.
+
+    Example:
+
+        import time
+        class Clock:
+            @classproperty
+            def sample():
+                return time.time()
+
+        # Different results each time, just like an instance method, but without having to instantiate the class.
+        Clock.sample
+        1665600812.008385
+        Clock.sample
+        1665600812.760394
+    """
+
+    def __init__(self, getter):
+        self.getter = getter
+
+    def __get__(self, instance, instance_class):
+        ignored(instance)
+        return self.getter(instance_class)
+
+
+class classproperty_cached(object):
+    """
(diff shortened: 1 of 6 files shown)


dcicutils/misc_utils.py
```diff
@@ -1560,6 +1567,110 @@ def singleton_class(self):
...
+
+    def __init__(self, getter):
+        self.getter = getter
+
+    def __get__(self, instance, instance_class):
+        ignored(instance)
+        return self.getter(instance_class)
+
+
+class classproperty_cached(object):
```
willronchetti, line 1597: You might consider providing a "reset" method that will "delete" the cached class property (so it is recomputed). I can see how this is useful but I also think under some situations you may want to trigger a discard of the cached value.

willronchetti (approved):
Just one small comment on `classproperty_cached`

PR #226, opened 2022-10-28
Elaborations of the classproperty decorators

netsettler:


dcicutils/misc_utils.py
```diff
@@ -1630,16 +1631,140 @@ class SubFreeze(Freeze):
...
+        predicate = issubclass if subclasses else equals
+        for cache_class, cache_value in attribute_cache.items():
+            if predicate(cache_class, reference_class):
+                keys_to_remove.append(cache_class)
+        for cache_class in keys_to_remove:
+            del attribute_cache[cache_class]
+        return bool(keys_to_remove)
+
+
+class classproperty_cached_each_subclass(classproperty_cached):
```
willronchetti, line 1724: Could you explain why this is named this way? Why not `classproperty_cached_first_use`?

dcicutils/misc_utils.py
```diff
@@ -1630,16 +1631,140 @@ class SubFreeze(Freeze):
...
+        1665600540.1467211
+
+        Freeze.sample
+        1665600374.4801269
+
+        SubFreeze.sample
+        1665600540.1467211
+    """
+
+    _USES_PER_SUBCLASS_CACHES = True
```
willronchetti, line 1760: Oh I see - I think the docstring for this class needs updating to reflect this. It looks the same as the parent class.

dcicutils/misc_utils.py
```diff
@@ -1630,16 +1631,140 @@ class SubFreeze(Freeze):
...
+
+
+def equals(x, y):
```
willronchetti, line 1763: If the only usage of this is in the class why not just do `lambda x, y: x == y` and avoid the function call/definition?

willronchetti (approved):

dcicutils/misc_utils.py
netsettler: If you have several subclasses of the indicated class, each will have its own value computed. None will share.

dcicutils/misc_utils.py
netsettler: Probably a copy error. Yeah, I'll fix that.

dcicutils/misc_utils.py
netsettler: I find these things useful from time to time and it annoys me coming from a language like Lisp that doesn't have an "operator" distinction not to have this available as a function, so it seemed that it would enable other uses prospectively in the future.

PR #227, opened 2022-11-09
Minor fix to override_dict for overriding os.environ value with non-str type.

dmichaels-harvard:
This came up in 4dn-cloud-infra where custom/config.json had a non-string value (e.g. `s3.bucket.encryption` set to `false`).


willronchetti (approved):
Just needs version bump

PR #229, opened 2022-12-05
OpenSearch + CB Overrides

willronchetti:
- Add equivalent OpenSearch client support
- Add `run_project_build_with_overrides` for use with pipeline images


netsettler (approved):
This looks good.

PR #231, opened 2023-01-03
Added env_equals to env_utils module.

dmichaels-harvard:


willronchetti (approved):
Might consider adding a test or two since testing this function should be relatively straightforward

dcicutils/env_utils.py
```diff
@@ -1026,6 +1026,10 @@ def foursight_env_name(envname):
     return infer_foursight_from_env(envname=envname)
 
 
+def env_equals(envname_a: str, envname_b: str) -> bool:
+    return foursight_env_name(envname_a) == foursight_env_name(envname_b)
```
netsettler, line 1030: It doesn't really matter if you compare this or full_env_name because _hopefully_ there are no names that are overlapping between the different spaces, but I might slightly prefer full_env_name here just because it's more primitive and perhaps a little faster to compute.

netsettler (approved):
My comment is not preemptive.

PR #236, opened 2023-01-06
Add keys_and_values_to_dict function to misc_utils

dmichaels-harvard:


CHANGELOG.rst
```diff
@@ -7,6 +7,14 @@ Change Log
 ----------
 
 
+6.6.0
+=====
+
+* In ``dcicutils.misc_utils``:
+
+  * Add ``keys_and_values_to_dict`` function (and associated unit test).
+
```
netsettler, line 16: I would not mind seeing a `dict_to_keys_and_values`, too.

dcicutils/misc_utils.py
```diff
@@ -1173,6 +1173,41 @@ def find_association(data, **kwargs):
...
+
+    :param keys_and_values: List of key/value objects as described above.
+    :param key_name: Name of the given key property in the given list of key/value objects; default to "Key".
+    :param value_name: Name of the given value property in the given list of key/value objects; default to "Value".
+    :returns: Dictionary of keys/values from given list of key/value object as described above.
+    """
+    result = {}
+    for item in keys_and_values:
+        key = item.get(key_name)
+        if key:
```
netsettler, line 1206: I might just do something like this (untested). I don't know that I like tolerating value_name being missing. And I think it's good to mention situations that are not in proper form. Do we expect these might be in bad form for the intended first use? In general, though, the point of putting something like this in a library is so it has multiple uses, so I think tolerating bad data is just not good unless you make an argument that requests it.
```
result = {}
for item in keys_and_values:
    if key_name not in item:
        raise ValueError(f"Key {key_name} is not in {item}.")
    if value_name not in item:
        raise ValueError(f"Key {value_name} is not in {item}.")
    if key_name in result:
        raise ValueError(f"Key {key_name} is duplicated in {keys_and_values}.")
    result[item[key_name]] = result[item[value_name]]
return result
```
Sorry if I told you in the other PR you'd get away with using the code you had there in that very specialized case. Now that it's here, new scrutiny required. :)

netsettler (approved):
Provisionally approved. If you feel strongly it shouldn't do error checking, let's talk more. But hopefully you can take the suggestions I've made, or something similar.

dcicutils/misc_utils.py
dmichaels-harvard: Yeah, sounds good, should we also check if the value of the key is a string (doesn't have to be I guess  - I was doing str on in just to be sure; and probably also check that given key_name and value_name are both (non-empty) strings.

CHANGELOG.rst
dmichaels-harvard: Good idea, all set.

dcicutils/misc_utils.py
dmichaels-harvard: All set.

dmichaels-harvard:
> Provisionally approved. If you feel strongly it shouldn't do error checking, let's talk more. But hopefully you can take the suggestions I've made, or something similar.

I took your suggestions, with error checking, thanks.

dcicutils/misc_utils.py
```diff
@@ -1199,12 +1199,57 @@ def keys_and_values_to_dict(keys_and_values: list, key_name: str = "Key", value_
...
+
+    :param keys_and_values: Dictionary of keys/values described above.
+    :param key_name: Name of the given key property in the result list of key/value objects; default to "Key".
+    :param value_name: Name of the given value property in the result list of key/value objects; default to "Value".
+    :returns: List of key/value objects from the given dictionary as described above.
+    """
+    result = []
+    for key in dictionary:
+        result.append({key_name: key, value_name: dictionary[key]})
     return result
```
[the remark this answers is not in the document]
dmichaels-harvard: Sounds good, updated, thanks.

PR #237, opened 2023-01-20
Auth0 Domain, Connections

willronchetti:
- Allow ini file manager to write values for Auth0Domain and Auth0AllowedConnections
- Tested with cgap-portal


dcicutils/deployment_utils.py
```diff
@@ -457,8 +461,10 @@ def build_ini_file_from_template(cls, template_file_name, init_file_name, *,
             tibanna_output_bucket (str): Specific name of the bucket to use on S3 for tibanna logs.
             application_bucket_prefix (str): An application bucket prefix to use, overriding the default one.
             foursight_bucket_prefix (str): A foursight bucket prefix to use, overriding the default one.
+            auth0_domain (str): A string identifying the Auth0 Domain to send auth requests to.
             auth0_client (str): A string identifying the auth0 client application.
             auth0_secret (str): A string secret that is passed with the auth0_client to authenticate that client.
+            auth0_allowed_connections (str): A comma separated string of allowed connections that can be used via auth0.
```
dmichaels-harvard, line 467: I might just add or change to mention this is supported _identity provider names_ or some such (e.g. google, github); at first thought this might be like the callback domains or something.

dmichaels-harvard (approved):
LGTM

dcicutils/deployment_utils.py
netsettler: And presumably any change in that wording gets copied into the other functions that take the same arg.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "dcicutils"
-version = "6.7.0"
+version = "6.8.0"
```
netsettler, line 3: This is presumably an incompatible change, not strictly a bug fix. I'd bump the major version, which would mean older versions of the system if something isn't upgraded wouldn't try to fetch incompatible support if someone did poetry update. That's the reason for making major version bumps.

netsettler (approved):
My comments aren't blocking.

PR #239, opened 2023-02-02
Added is_c4_arn function to misc_utils.

dmichaels-harvard:


test/test_misc_utils.py
```diff
@@ -1419,7 +1419,7 @@ def test_check_true():
     msg = "x is not a list of four, five, and six."
     with pytest.raises(RuntimeError) as e:
         check_true(x == [4, 5, 6], msg)
-    assert msg in str(e)
+        assert msg in str(e)
```
dmichaels-harvard, line 1422: This seems to have been misplaced, test was failing.

netsettler: No, it's not misplaced. The assert is passing now because it is never reached. Something else is wrong with the test.

dmichaels-harvard: Oooh. Lemme see.

dmichaels-harvard: Fixed with your suggestion, changing str(e) to str(e.value). Thanks.

CHANGELOG.rst
```diff
@@ -7,6 +7,14 @@ Change Log
 ----------
 
 
+6.7.1
```
netsettler, line 10: New functionality is a minor version change. We're at 6.8 so you'll need to be 6.9.0.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "dcicutils"
-version = "6.7.0"
+version = "6.7.1"
```
netsettler, line 3: 6.9

test/test_misc_utils.py
```diff
@@ -3159,6 +3159,16 @@ def test_dict_to_keys_and_values():
     assert dict_to_keys_and_values({}) == []
 
 
+def test_is_c4_arn():
+    assert is_c4_arn('arn:aws:ecs:us-east-1:643:cluster/c4-ecs-cgap-production-stack-Productionblue-fzOm82DGZKGZ')
+    assert is_c4_arn('arn:aws:ecs:us-east-1:643:cluster/c4-ecs-cgap-production-stack-CgapProductionblue-fzOm82DGZKGZ')
+    assert is_c4_arn('arn:aws:ecs:us-east-1:643:cluster/c4-ecs-fourfront-hotseat-stack-Hotseat-ZnIAAAz8Apxf')
+    assert is_c4_arn('arn:aws:ecs:us-east-1:643:cluster/c4-ecs-fourfront-production-stack-FourfrontProd-hU0JiwOeWfdm')
+    assert not is_c4_arn('arn:aws:ecs:us-east-1:643:cluster/c4-ecs-abc-webdev-stack-CgapWebdev-rSLwZBbdVTtx')
```
netsettler, line 3167: Perhaps we should use a regexp match and look for "[:/]c4-" in the name as well?

test/test_misc_utils.py
```diff
@@ -3159,6 +3159,16 @@ def test_dict_to_keys_and_values():
...
+    assert not is_c4_arn('arn:aws:ecs:us-east-1:643:cluster/default')
+    assert not is_c4_arn('arn:aws:ecs:us-east-1:643:cluster/c4-ecs-def-mastertest-stack-FourfrontMaster-O25WVcRMpExr')
```
netsettler, line 3169: Probably picking some completely different ARN is worth testing, too, though I guess /default thing above is not really that different. I'd put that test case last if you make the c4- things return true, just for grouping.

netsettler (approved):
I think we should add the c4- test I mentioned in comments.
Approved if you make that change.

CHANGELOG.rst
dmichaels-harvard: Got it.

test/test_misc_utils.py
dmichaels-harvard: You mean in the `is_c4_arn` function?

netsettler: (and doing re.search, not re.match)

dmichaels-harvard: Got it, thanks.

PR #240, opened 2023-02-13
Redis Utils

willronchetti:
- Creates 2 new files `redis_utils` and `redis_tools`, the former contains low level APIs for interacting with Redis as a basic key/value store and the latter contains a higher level interface for managing sessions with Redis
- Remove unused `pytest.ini` file
- Update/add dependencies for Redis
- Some basic tests


dcicutils/redis_utils.py
```diff
@@ -0,0 +1,89 @@
...
+              sub classes later on to isolate in a reasonable manner, but for now all here
+    """
+
+    def __init__(self, redis_handle):
+        self.redis = redis_handle
+
+    @staticmethod
+    def _encode_value(value):
+        """ Function used to preprocess all values that are strings and encode them as utf-8 bytes """
+        return value.encode('utf-8') if (value and isinstance(value, str)) else value
```
netsettler, line 41: So this is not needed for other types? Or those other types are not given? Something about the expected args would be helpful in the doc string. I was assuming you'd just `json.dumps(value).encode('utf-8')` to do the encoding and `json.loads(value.decode('utf-8'))` to pull the value back, but maybe you're trying to be more efficient than that. If so, should you perhaps error check the argument type, or, minimally, declare the argument type you expect (which has only informational and not runtime effect)?

pyproject.toml
```diff
@@ -68,25 +69,28 @@ flaky = ">=3.7.0"
 pytest = ">=4.5.0"
 pytest-cov = ">=2.7.1"
 pytest-mock = ">=1.11.0"
+pytest-redis = "^2.0.0"
```
netsettler, line 72: I'll be curious what this contributes.

netsettler (approved):
Yeah, I don't see a blocker. Obviously, I've asked some design questions but if you're OK with your choices in the face of those questions, I think we can go with it.

dmichaels-harvard (approved):

dmichaels-harvard (approved):
LGTM

dcicutils/secrets_utils.py
```diff
@@ -117,7 +117,8 @@ def apply_overrides(*, secrets: dict, rename_keys: Optional[dict] = None,
                 raise ValueError(f"Cannot rename {key} to {new_name}"
                                  f" because {a_or_an(new_name)} attribute already exists.")
             if key not in secrets:
-                raise ValueError(f"Cannot rename {key} to {new_name} because the secrets has no {key} attribute.")
+                logger.warning(f"Cannot rename {key} to {new_name} because the secrets has no {key} attribute.")
```
netsettler, line 120: New situation needs new unit tests.

netsettler (approved):
My comment is not blocking.

test/test_redis_utils.py
```diff
@@ -66,3 +68,23 @@ def build_item_metadata():
...
+        rd = RedisBase(redisdb)
+        my_key_meta = 'snovault:items:uuid:meta'
+        rd.set(my_key_meta, 'hello')
+        assert rd.get(my_key_meta) == 'hello'
+        rd.delete(my_key_meta)
+        # test exp with datetime
+        exp = datetime.timedelta(seconds=2)
+        rd.set(my_key_meta, 'hello', exp=exp)
+        assert rd.get(my_key_meta) == 'hello'
+        time.sleep(3)
```
netsettler, line 83: Is it possible to be using ControlledTime in this test to avoid having to have tests take a long time?

netsettler: I spoke to you on Slack, and you said not. Oh well.

netsettler (approved):
I already approved, but I've checked recent changes and still do.

dcicutils/redis_utils.py
```diff
@@ -59,13 +60,17 @@ def info(self) -> dict:
...
         :param key: string to store value under
         :param value: value mapped to from key
+        :param exp: expiration time in seconds as int or datetime.timedelta (optional)
         :return: a "true" string <OK>
         """
-        return self.redis.set(self._encode_value(key), self._encode_value(value))
+        kwargs = {}
+        if exp:
+            kwargs['ex'] = exp
+        return self.redis.set(self._encode_value(key), self._encode_value(value), **kwargs)
```
dmichaels-harvard, line 73: Why using kwargs here instead of just `ex=exp` in redis.set call? (Nothing wrong with that just wondering).

willronchetti: Just future proofing it in case additional kwargs are needed in the future

PR #243, opened 2023-03-01
Small UnifiedAuthenticator refactor and other small changes

netsettler:
This consolidates a bunch of smallish fixes I'd been accumulating on various branches for a while.

### DIFF
--- a/dcicutils/ff_utils.py
+++ b/dcicutils/ff_utils.py
@@ -242,9 +242,8 @@ def internal_compute_authorized_request(url, *, auth, ff_env, verb, retry_fxn, *
     if 'timeout' not in kwargs:
         kwargs['timeout'] = 60  # default timeout
 
-    verb_upper = verb
+    verb_upper = verb.upper()
     try:
-        verb_upper = verb.upper()
         the_verb = REQUESTS_VERBS[verb_upper]
     except KeyError:
         raise ValueError(f"Provided verb {verb} is not valid. Must be one of {disjoined_list(REQUESTS_VERBS)}.")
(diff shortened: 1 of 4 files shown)


dcicutils/ff_utils.py
```diff
@@ -1211,35 +1210,76 @@ def search_es_metadata(index, query, key=None, ff_env=None, is_generator=False):
...
+    @classmethod
+    def maybe_unwrap_legacy_auth(cls, auth):
+        # Compatibility with old form of auth from get_key()
+        # If {"default": {...auth...}, ...} is given, peel off outer wrapper and use the inner {...auth...} part.
+        if isinstance(auth, dict) and isinstance(auth.get('default'), dict):
+            return auth['default']
+        return auth
+
+    @classmethod
+    def normalize_auth(cls, auth):
```
dmichaels-harvard, line 1272: Do we generally regard it as too tedious to use argument and return value hints? They'd be somewhat useful for me anyways, e.g. normalize_auth(cls, auth: Optional[Union[dict, Tuple]) -> Optional[Tuple]:

willronchetti: I'm good with adding type hints so long as we're here. I do try to add them in new code nowadays.

aschroed (approved):
I guess you have a good reason for this and still looks to support the same functionality so OK by me, FWIW.

dmichaels-harvard (approved):

dcicutils/s3_utils.py
```diff
@@ -154,6 +161,9 @@ def __init__(self, outfile_bucket=None, sys_bucket=None, raw_file_bucket=None,
                     tibanna_output_bucket = tibanna_output_bucket_from_health_page
                 logger.warning('Buckets resolved successfully.')
             else:
+                # TODO: We require global env bucket to be set everywhere now, so can we throw an error here?
+                #       -kmp 1-Mar-2023
```
willronchetti, line 165: I think throwing an error is fine at this point.

dcicutils/s3_utils.py
```diff
@@ -260,52 +271,73 @@ def health_page_get(self, health_page_key: str, *, default: Any = None, force: b
 
     ACCESS_KEYS_S3_KEY = 'access_key_admin'
 
-    def get_access_keys(self, name=ACCESS_KEYS_S3_KEY):
+    def get_access_keys(self, name: str = ACCESS_KEYS_S3_KEY) -> ServerAuthDict:
+        # TODO: Make .get_access_keys() do better error checking?
+        #       I have marked the return value as ServerAuthDict because it's CLEAR that various callers
+        #       assume that values coming back from this function are in the right form. e.g.,
+        #       ff_utils.get_authentication_with_server will give a hard error if this ever returns
+        #       something with a server missing.  So presumably we mostly do the right thing here.
```
willronchetti, line 279: These access key related methods are all operating under the assumption that these S3 files follow the convention that `load_access_keys` uses ie:

https://github.com/4dn-dcic/fourfront/blob/master/src/encoded/commands/load_access_keys.py#L77-L79
https://github.com/dbmi-bgm/cgap-portal/blob/master/src/encoded/commands/load_access_keys.py#L68-L70

dcicutils/s3_utils.py
```diff
@@ -260,52 +271,73 @@ def health_page_get(self, health_page_key: str, *, default: Any = None, force: b
...
+        akey: Union[bytes, str] = response['Body'].read()
         if type(akey) == bytes:
-            akey = akey.decode()
+            akey: str = akey.decode()
         try:
             return json.loads(akey)
         except (ValueError, TypeError):
+            # TODO: Is there really a use case for this returning non-JSON data? If so, can we make a different
+            #       function that does this? Because this is just going to return garbage when peopla are expecting
+            #       key dictionaries and we really should raise an error in that case. -kmp 1-Mar-2023
```
willronchetti, line 331: Yes there is, in the case of Higlass or JH keys.

dcicutils/s3_utils.py
```diff
@@ -260,52 +271,73 @@ def health_page_get(self, health_page_key: str, *, default: Any = None, force: b
...
 
         if isinstance(keys.get('default'), dict):
             keys = keys['default']
         if self.url:
+            # NOTE WELL: self.url might contain the empty string if S3Utils was initialized with a system bucket
+            #            instead of an ff_env. Moreover, we are overwriting whatever server was actually stored
+            #            on the server, so we might be overwriting good data with bad. We should document that in
+            #            fact self.url needs to always be set to a non-empty value by the time we reach here, and
+            #            probably raise an error otherwise. We should also raise an error if there's a server
+            #            in the data we retrieved because it's just not going to get used.  -kmp 1-Mar-2023
```
willronchetti, line 294: So long as envs do not share buckets it would be safe to do such a thing but I'd advise against it simply because the current protocol will always put a server up.

dcicutils/s3_utils.py
```diff
@@ -260,52 +271,73 @@ def health_page_get(self, health_page_key: str, *, default: Any = None, force: b
...
 
-    def get_access_keys(self, name=ACCESS_KEYS_S3_KEY):
+    def get_access_keys(self, name: str = ACCESS_KEYS_S3_KEY) -> ServerAuthDict:
+        # TODO: Make .get_access_keys() do better error checking?
+        #       I have marked the return value as ServerAuthDict because it's CLEAR that various callers
+        #       assume that values coming back from this function are in the right form. e.g.,
+        #       ff_utils.get_authentication_with_server will give a hard error if this ever returns
+        #       something with a server missing.  So presumably we mostly do the right thing here.
+        #       It would be much better to properly error-check the result here or even in .get_key().
+        #       -kmp 1-Mar-2023
```
willronchetti, line 281: If you want to error check do it here instead of in `get_key` which does need to be able to decode/decrypt either JSON or plaintext data.

dcicutils/s3_utils.py
```diff
@@ -260,52 +271,73 @@ def health_page_get(self, health_page_key: str, *, default: Any = None, force: b
...
+            #            fact self.url needs to always be set to a non-empty value by the time we reach here, and
+            #            probably raise an error otherwise. We should also raise an error if there's a server
+            #            in the data we retrieved because it's just not going to get used.  -kmp 1-Mar-2023
             keys['server'] = self.url
-        return keys
+        return keys  # Might still have no server if self.url was not a URL
 
-    def get_ff_key(self):
+    def get_ff_key(self) -> ServerAuthDict:
+        # TODO: The return value should be just AuthDict if we decide get_access_keys() can't guarantee ServerAuthDict.
```
willronchetti, line 299: Until we remove support for Admin keys (as opposed to Foursight or Tibanna keys), this value should always be available, otherwise the system has not been orchestrated correctly (no deployment).

dcicutils/s3_utils.py
```diff
@@ -260,52 +271,73 @@ def health_page_get(self, health_page_key: str, *, default: Any = None, force: b
...
         return self.get_access_keys()
 
-    def get_higlass_key(self):
+    def get_higlass_key(self) -> Union[AnyJsonData, str]:  # parsed json or string that won't parse
+        # TODO: It would be nice if this could be returning something with a better promise. -kmp 1-Mar-2023
```
willronchetti, line 303: This actually will not return JSON data but raw text I believe. Same deal with Google/JH keys. I don't actually think the Google key is used anymore.

dcicutils/s3_utils.py
netsettler: OK, I'm preparing a layered PR that will include error-checking of the form you're approving here. This has been added.

dcicutils/s3_utils.py
netsettler: `access_key_foursight` follows similar patterns. No evidence that other than a dictionary is expected.

dcicutils/s3_utils.py
```diff
@@ -410,15 +442,16 @@ def get_file_size(self, key, bucket=None, add_bytes=0, add_gb=0,
...
         if not bucket:
             bucket = self.outfile_bucket
         self.s3.delete_object(Bucket=bucket, Key=key)
 
     @classmethod
-    def size(cls, bucket):
+    def size(cls, bucket: str) -> int:
         sbuck = boto3.resource('s3').Bucket(bucket)
         # get only head of objects so we can count them
+        # Is there really no .count() ? -kmp 1-Mar-2023
```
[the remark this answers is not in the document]
netsettler: OK, I updated the doc a little, and cross-checked this claim against StackOverflow. Apparently there is a page in the billing part of the AWS console that knows the answer to this, but to do it programmatically (without scraping that page), this is the best that could be done as of time of that page.

dcicutils/s3_utils.py
netsettler: I'm not sure this is still needed, but we can see about making a change another time. I added some notes on another comment a little farther down this same page.

dcicutils/s3_utils.py
netsettler: I still kinda think, after doing more research, that the plaintext data is no longer needed and that we just have unit tests that are keeping the old path alive. But not going to obsess about that now. We can circle back for that.

dmichaels-harvard (approved):

dcicutils/s3_utils.py
netsettler: I think I've dealt with this. My concern was not in changing the requirement but in understanding the requirement. From the existing code at time I started, the part you say you were relying on could not be statically proven.

PR #247, opened 2023-03-31
Add trace_utils and fix obfuscation_utils

netsettler:
* New `trace_utils` module

  * New decorator `@Trace`

  * New function `make_trace_decorator` to make similar ones.

* Fix to `obfuscation_utils` relating to dicts containing lists.

### DIFF
--- a/dcicutils/trace_utils.py
+++ b/dcicutils/trace_utils.py
@@ -1,0 +1,50 @@
+import functools
+import io
+
+from dcicutils.misc_utils import environ_bool, full_object_name, get_error_message, PRINT
+from dcicutils.obfuscation_utils import obfuscate_dict
+
+
+def _expand_dict(d, indent=0):
+    s = io.StringIO()
+    d = obfuscate_dict(d)
+    if not d:
+        PRINT("dict()", file=s)
+    else:
+        PRINT("dict(", file=s)
+        for k, v in d.items():
+            PRINT(f"{' ' * (indent + 2)}{k}={v!r},", file=s)
+        PRINT(f"{' ' * indent})", file=s)
+    return s.getvalue()
+
+
+def make_trace_decorator(enabled_by_default=False):
+
+    def _trace_decorator(enabled=None):
+        if enabled is None:
+            enabled = enabled_by_default or environ_bool('TRACE_ENABLED')
+
+        def _maybe_attach_trace(fn):
+            if not enabled:
+                return fn
+            trace_name = full_object_name(fn)
+
+            @functools.wraps(fn)
+            def _traced(*args, **kwargs):
+                PRINT(f"Entering {trace_name} with args={args!r} kwargs={_expand_dict(kwargs)}")
+                try:
+                    res = fn(*args, **kwargs)
+                    PRINT(f"Function {trace_name} returned {res!r}")
+                    return res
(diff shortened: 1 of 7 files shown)


dcicutils/trace_utils.py
```diff
@@ -0,0 +1,56 @@
...
+            enabled = enabled_by_default or environ_bool('TRACE_ENABLED')
+
+        def _maybe_attach_trace(fn):
+            if not enabled:
+                return fn
+            trace_name = full_object_name(fn)
+
+            @functools.wraps(fn)
+            def _traced(*args, **kwargs):
+                PRINT(f"Entering {trace_name} with args={_maybe_obfuscate(args)!r} kwargs={_maybe_obfuscate(kwargs)!r}")
```
dmichaels-harvard, line 40: Totally just style but for stuff like this I prefer to see caps, e.g. ```PRINT("TRACE {trace_name} | ARGS: {_maybe_obfuscate(args)!r} | KWARGS: {_maybe_obfuscate(kwargs)!r}"```

dcicutils/obfuscation_utils.py
```diff
@@ -115,6 +119,14 @@ def has_values_to_obfuscate(dictionary: dict) -> bool:
     for key, value in dictionary.items():
         if isinstance(value, dict):
             dictionary[key] = obfuscate_dict(value, show=False, inplace=False, obfuscated=obfuscated)
+        elif isinstance(value, list):
```
dmichaels-harvard, line 122: I guess you picked this up from my branch (unless you somehow made precisely the same change :) 
Question: This also does not handle passing in a list at top-level - do you think it should? Of course, the function name (obfuscate_dict) would be off, but would probably good to handle that too. If so I can take care of that (and new tests).

dcicutils/trace_utils.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+# You can use TRACE_REDACT=FALSE if tracing should show credentials,
+# but obviously this is usually not something you'd want to do, so this
+# is on a separate variable that defaults to True. -kmp 30-Mar-2023
+
+TRACE_REDACT = environ_bool("TRACE_REDACT", default=True)
+
+
+def _maybe_obfuscate(x):
```
dmichaels-harvard, line 14: Note in the case of SubmitCGAP this will print the value of the access secret in plaintext, as part of the `auth` tuple. Not sure we want a special case for this.

dcicutils/trace_utils.py
```diff
@@ -0,0 +1,56 @@
...
+            if not enabled:
+                return fn
+            trace_name = full_object_name(fn)
+
+            @functools.wraps(fn)
+            def _traced(*args, **kwargs):
+                PRINT(f"Entering {trace_name} with args={_maybe_obfuscate(args)!r} kwargs={_maybe_obfuscate(kwargs)!r}")
+                try:
+                    res = fn(*args, **kwargs)
+                    PRINT(f"Function {trace_name} returned {_maybe_obfuscate(res)!r}")
```
dmichaels-harvard, line 43: Maybe special case, as I'd done in mine, for displaying the `json()` of a `requests.model.Request`; I know it's kinda lame having such random special cases, but can be useful.

PR #248, opened 2023-03-31
Add trace utils and obfuscation fix to other misc changes

netsettler:
This was originally a layered PR, adding support to [branch kmp_misc_20230228](https://github.com/4dn-dcic/utils/tree/kmp_misc_20230228) ([PR #243](https://github.com/4dn-dcic/utils/pull/243)), but I've rebased this to master because those other parts are already in there.

* New ``trace_utils`` module ([PR #247](https://github.com/4dn-dcic/utils/pull/247))

  * New decorator ``@Trace``

  * New function ``make_trace_decorator`` to make similar ones.

* Fix to ``obfuscation_utils`` relating to dicts containing lists. ([PR #246](https://github.com/4dn-dcic/utils/pull/246))

* New function `misc_utils.deduplicate_list`.

* Fixes to the `printed_output`  context manager in `qa_utils`.


willronchetti (approved):
Looks very similar to our `debuglog` facility but I think still useful for tracking function use generally.

netsettler:
> Looks very similar to our `debuglog` facility

No. `debuglog` lets you put in your own messages and is a way of causing them to go to a file rather than normal logging stuff for quicky debugging that doesn't clutter cloudtrail.

This doesn't take a position on where the output goes (though maybe it should be like `debuglog`, we could think about that. Right now it goes to where PRINT would go. But the purpose of this is so you don't have to explicitly insert `debuglog `commands into the code. It just figures out what you'd want to print. So that's pretty different.

netsettler:
I'm adding `deduplicate_list`, written by Will in `snovault`.

dcicutils/obfuscation_utils.py
```diff
@@ -96,25 +154,30 @@ def obfuscate_dict(dictionary: dict, inplace: bool = False, show: bool = False,
...
+        return item
+
+    orig_item = item
+
+    changed = StorageCell(False)
+
+    if not inplace:
+        item = copy.deepcopy(item)
+
+    def process_recursively(item):
```
dmichaels-harvard, line 167: I see you took out has_values_to_obfuscate - I think that's a good call, was thinking about that, not worth the extra effort/processing-time.

dcicutils/obfuscation_utils.py
```diff
@@ -96,25 +154,30 @@ def obfuscate_dict(dictionary: dict, inplace: bool = False, show: bool = False,
...
-                    return True
-            elif isinstance(value, str) and should_obfuscate(key) and not is_obfuscated(value):
-                return True
-        return False
+    if show:
+        return item
+
+    orig_item = item
+
+    changed = StorageCell(False)
```
dmichaels-harvard, line 162: Oh, weird, I've only just not learned (sort of) why this `changed` variable needs to be a class/object; not intuitive to me; so when `process_recursively` is called recursively, somehow if it is a value (non-reference) type (like just bool) then it would be a new instance of it; could maybe use some good Kent-style pedagogy on this. (ChatGPT is trying to explain it, and experimenting with it, but still not 100% there in my understanding).

dcicutils/obfuscation_utils.py
```diff
@@ -75,7 +76,64 @@ def is_obfuscated(value: str) -> bool:
     return isinstance(value, str) and bool(OBFUSCATED_VALUE.match(value))
 
 
-def obfuscate_dict(dictionary: dict, inplace: bool = False, show: bool = False,
+# def obfuscate_dict(dictionary: dict, inplace: bool = False, show: bool = False,
```
dmichaels-harvard, line 79: Did you say you wanted to rename to `obfuscate_json` ?

dcicutils/obfuscation_utils.py
```diff
@@ -96,25 +154,30 @@ def obfuscate_dict(dictionary: dict, inplace: bool = False, show: bool = False,
...
+
+    if not inplace:
+        item = copy.deepcopy(item)
+
+    def process_recursively(item):
+        # We only need to process non-atomic items recursively, since they are the only things
+        # that might conceivably be or contain a dictionary in need of obfuscation.
+        if isinstance(item, dict):
+            for key, value in item.items():
+                if should_obfuscate(key):
```
dmichaels-harvard, line 172: I think we might want this to avoid obfuscating values which appear already to be obfuscated: ```if should_obfuscate(key) and not is_obfuscated(value):```

dmichaels-harvard: And actually, technically, since we can pass in a specific `obfuscated` value to use, the `is_obfuscated` function should also check if the given value is equal to this given `obfuscated` value, no?

dcicutils/obfuscation_utils.py
netsettler: The problem is that in any function, if you assign a variable, that variable becomes local to the function. You can read the value of an outer variable, but you cannot write one. (In fact, in some version of Python, I'm not sure which, there is a way of declaring (similar to `global`) that you mean to share the variables between the contexts so the assignment should happen farther out, but since `dcicutils` spans several versions, I didn't do that. And in any case this StorageCell thing can be used for many such things.

No, it is not creating a new instance on recursive calls. The only instance creation is outside the function, so it is a shared single location that all locations see and asks the question, "was there any changed value at all?"  I could, alternatively, just always copy. But this satisfies the previous unit tests, most of them.  (There is also a way to efficiently avoid having to do this which I can show you, but it was late and I was tired so I did not use that technique because I've never done it in Python and I'd have to puzzle it out a little more than I would in Lisp. I did change the unit test so that if something "looked obfuscated", it still ends up thinking it's copied it when perhaps it hasn't.)

dcicutils/obfuscation_utils.py
```diff
@@ -96,25 +154,30 @@ def obfuscate_dict(dictionary: dict, inplace: bool = False, show: bool = False,
...
+                    changed.value = True
+                    item[key] = obfuscate(value, obfuscated=obfuscated)
+                else:
+                    process_recursively(value)
+        elif isinstance(item, list):
```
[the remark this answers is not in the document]
netsettler: I guess you didn't have a unit test for that. :)  Unit tests are responsible for noticing that you've changed something of consequence during a refactor. It's one of the primary reasons I'm so manic about them.

I'd want to think on this more. Do we know a use case/ It's semantically perilous to do an in-place change of the sort you mention because tuples are considered constant and you'd be changing the hash value of a constant. I suppose  since it's always newly made structure, no ill is going to come of that. I'll experiment with that if you think it's going to break something.  In practice, I'm not sure we ever store tuples in this way.

dcicutils/obfuscation_utils.py
netsettler: The "compatible" part (and it's not quite functionally compatible, as you note for tuples, is the synonym definition at the bottom of the file. It's call-compatible.

dcicutils/obfuscation_utils.py
```diff
@@ -75,7 +76,64 @@ def is_obfuscated(value: str) -> bool:
...
+#                     obfuscated_value.append(obfuscate_dict(item, show=False, inplace=False, obfuscated=obfuscated))
+#                 else:
+#                     obfuscated_value.append(item)
+#             dictionary[key] = obfuscated_value
+#         elif isinstance(value, str) and should_obfuscate(key) and not is_obfuscated(value):
+#             dictionary[key] = obfuscate(value, show=False, obfuscated=obfuscated)
+#     return dictionary
+
+
+def obfuscate_dict(item: AnyJsonData, inplace: bool = False, show: bool = False,
```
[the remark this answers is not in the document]
netsettler: I wonder if there's a way of saying "returns the type of the input". I'll have to check on that. But yes, I suppose so. The `tuple` thing would add technical complexity to that.

dcicutils/obfuscation_utils.py
netsettler: > `if should_obfuscate(key) and not is_obfuscated(value):`

Again, there were no unit tests covering this case. But isn't this a job for `should_obfuscate`?

dcicutils/obfuscation_utils.py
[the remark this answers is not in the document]
dmichaels-harvard: My branch (obfuscation-utils-fix-20230330) had a couple test for tuples ... ```def test_obfuscate_dict_with_nested_tuple():

    d = {"abc": "123", "def": ({"ghi": "456", "jklsecret": "obfuscatethisvalue"}, 789), "jkl": "hello"}
    o = {"abc": "123", "def": ({"ghi": "456", "jklsecret": "<REDACTED>"}, 789), "jkl": "hello"}
    d_copy = copy.deepcopy(d)  # to make sure the original is not inadvertantly modified
    x = obfuscate_dict(d, obfuscated="<REDACTED>")
    assert x == o
    assert d == d_copy


def test_obfuscate_dict_with_tuple():

    d = ({"abc": "123", "def": ({"ghi": "456", "jklsecret": "obfuscatethisvalue"}, 789), "passwd": "hello"}, (1, 2, 3))
    o = ({"abc": "123", "def": ({"ghi": "456", "jklsecret": "<REDACTED>"}, 789), "passwd": "<REDACTED>"}, (1, 2, 3))
    d_copy = copy.deepcopy(d)  # to make sure the original is not inadvertantly modified
    x = obfuscate_dict(d, obfuscated="<REDACTED>")
    assert x == o
    assert d == d_copy```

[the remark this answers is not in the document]
dmichaels-harvard: Tuples do come up the TRACE_ENABLED=true of submit-ontology (the auth tuple is output, secret in plain text).

dcicutils/obfuscation_utils.py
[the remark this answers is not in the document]
dmichaels-harvard: Actually, return hint (and arg) might be better being just `Any` as any type can be passed (and returned) - if not a dict or list it just returns the value, e.g. obfuscate_dict(123) == 123

dcicutils/obfuscation_utils.py
dmichaels-harvard: Thanks for that. Looks like could also use the `nonlocal` keyword feature.

dcicutils/obfuscation_utils.py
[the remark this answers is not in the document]
netsettler: OK, I'm adding these tests and adjusting the type signature.

dcicutils/obfuscation_utils.py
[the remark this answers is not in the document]
netsettler: Using `Union[AnyJsonData, tuple]` seems to satisfy PyCharm.

PR #249, opened 2023-04-14
GlacierUtils

willronchetti:
- Adds `glacier_utils.py`, a library for interacting with/restoring objects from Glacier storage
- 4 phase restore process
  - Issue restore API call
  - Issue copy API call
  - Patch portal metadata
  - Issue delete markers to Glacier objects (optional if restore is temporary)


dcicutils/glacier_utils.py
```diff
@@ -0,0 +1,321 @@
...
+            resp = self.restore_portal_from_glacier(atid, days=days)
+            if resp:
+                success.append(atid)
+            else:
+                errors.append(atid)
+        if len(errors) == 0:
+            PRINT(f'Errors encountered restoring @ids: {errors}')
+        else:
+            PRINT(f'Successfully triggered restore requests for all @ids passed {success}')
+        return success, errors
```
netsettler, line 246: This kind of passive result strategy is not my personal preference. I'd rather just see success communicated as normal return and errors by raising. If the issue is raising multiple errors, see the `MultiError` class that I wrote in my [PR #250](https://github.com/4dn-dcic/utils/pull/250).

dcicutils/glacier_utils.py
```diff
@@ -0,0 +1,321 @@
...
+
+class GlacierRestoreException(Exception):
+    pass
+
+
+class GlacierUtils:
+
+    def __init__(self, env_name: str):
+        """ Pass an env_name that exists in your ~/.cgap_keys.json or ~/.fourfront_keys.json file
+            No support for admin keys!
```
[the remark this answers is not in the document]
willronchetti: I mean the admin keys you can acquire by using `s3Utils`

dcicutils/glacier_utils.py
```diff
@@ -0,0 +1,321 @@
...
+            # Get temporary location
+            restore = self.extract_temporary_s3_location_from_restore_response(bucket, key)
+            # Copy the object from the temporary location to its original location
+            copy_source = {'Bucket': bucket, 'Key': key, 'Restore': restore}
+            copy_args = {'Bucket': bucket, 'Key': key}
+            if version_id:
+                copy_source['VersionId'] = version_id
+                copy_args['CopySourceVersionId'] = version_id
+            response = self.s3.copy_object(CopySource=copy_source, **copy_args)
+            PRINT(f'Response from boto3 copy:\n{response}')
```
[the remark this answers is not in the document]
willronchetti: I would like this to show up both when a user is running it from the console and when run programmatically say within foursight. I _think_ PRINT accomplishes both?

dcicutils/glacier_utils.py
willronchetti: My thinking is the caller should handle errors as needed? I want something that won't crash mid run. Having this go part way can be dangerous - special care is needed when using this.

dcicutils/glacier_utils.py
```diff
@@ -0,0 +1,321 @@
...
+from tqdm import tqdm
+from .misc_utils import PRINT
+from .ff_utils import get_metadata, get_health_page, patch_metadata
+from .creds_utils import KeyManager
+
+
+GLACIER_CLASSES = [
+    'GLACIER',
+    'DEEP_ARCHIVE'
+]
```
drio18, line 13: I'm guessing `GLACIER` here refers to what is now Glacier Flexible Retrieval per documentation here: https://aws.amazon.com/s3/storage-classes/. I don't know if that value still corresponds to Glacier Flexible Retrieval, or covers both Glacier Flexible Retrieval and Glacier Instant Retrieval now. If you're confident on the topic, then consider adding a quick comment for clarification. If not, consider reducing to only `DEEP_ARCHIVE` (since we've only ever used Glacier Deep Archive as far as I know), checking all current Glacier storage classes are covered with these 2 options, or adding a comment to look more into this later.

dcicutils/glacier_utils.py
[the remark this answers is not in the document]
netsettler: Well, it's just doing I/O to stdout by default (you can override with `file=` as in `print`), so yeah, if that's the normal way that it's done for other `foursight` things, I guess that's right. OK, just wanted to check.

That kind of gets to another question I had, which was who was the likely caller. These are potentially long-running things, so I assume it would not work well to do this kind of thing in a web request, for example.

dcicutils/glacier_utils.py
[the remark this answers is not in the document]
netsettler: Ah. I might have said `shared admin keys` there. But then again the real question is whether you want this remark in the source or in the doc. Doc strings will end up on readthedocs. I guess that comment will, for a time, be helpful to readers who've been using things an old way. After a while, the typical reader may think you mean something else, like an aspiration not achieved.

dcicutils/glacier_utils.py
```diff
@@ -0,0 +1,321 @@
...
+                        VersionId=v.get('VersionId')
+                    )
+                    PRINT(f'Object {key} Glacier version {v.get("VersionId")} deleted:\n{response}')
+                # no Glacier versions were found
+                if not glacier_versions:
+                    PRINT(f'No Glacier versions found for object {bucket}/{key}')
+                    return False
+                else:
+                    return True
+            else:
```
drio18, line 183: This seems to be missing appropriate return values.

dcicutils/glacier_utils.py
```diff
@@ -0,0 +1,321 @@
...
+        :return: True if success or False if failed
+        """
+        try:
+            # Retrieve the object versions for the key
+            response = self.s3.list_object_versions(Bucket=bucket, Prefix=key)
+            versions = response.get('Versions', [])
+            versions.sort(key=lambda x: x['VersionId'], reverse=True)  # most recent version will be first
+
+            # Delete all glaciered versions if the flag is set
+            if delete_all_versions:
```
drio18, line 168: Might be missing something here, isn't the same thing happening here regardless of the value of `delete_all_version`? For all versions, the storage class is checked, then the object is deleted and lines are logged.

dcicutils/glacier_utils.py
```diff
@@ -0,0 +1,321 @@
...
+            restore_path = response.get('RestoreOutputPath')
+            if restore_path is None:
+                PRINT(f'Error: Could not determine the temporary location of restored object {bucket}/{key}')
+                return None
+            return restore
+        except Exception as e:
+            PRINT(f'Error copying object {bucket}/{key} back to its original location in S3: {str(e)}')
+            return None
+
+    def patch_file_lifecycle_status_to_standard(self, atid: str, status: str,
```
drio18, line 130: Consider making `status` have a default of `uploaded` here, since that's correct for the vast majority of situations.

drio18 (commented):
Generally looks good; leaving a couple clarifying comments and suggestions, but nothing that will block merging (besides adding some useful tests if possible with aws mocks). I think this is a great start as something to run from personal command lines for now, and once we're confident things are looking good, we can move it to foursight potentially depending on the lamdba timeout issues you previously mentioned.

dcicutils/glacier_utils.py
willronchetti: I've cleaned up this method considerably to address this comment and the above, thanks!

test/test_glacier_utils.py
```diff
@@ -0,0 +1,271 @@
...
+    ])
+    def test_glacier_utils_delete_glaciered_versions_exist(self, glacier_utils, response):
+        """ Tests some mocked deletes of dummy list_objects calls for a particular object with a standard
+            and glacier version, should return true """
+        gu = glacier_utils
+        with mock.patch.object(gu.s3, 'list_object_versions',
+                               return_value=response):
+            with mock.patch.object(gu.s3, 'delete_object'):
+                assert gu.delete_glaciered_object_versions('bucket', 'key')
+                assert gu.delete_glaciered_object_versions('bucket', 'key', delete_all_versions=True)
```
netsettler, line 223: After doing these calls, you should be testing the calls really occurred. e.g.,

```
            with mock.patch.object(gu.s3, 'delete_object') as mock_delete:
                assert gu.delete_glaciered_object_versions('bucket', 'key')
                assert mock_delete.call_count == 1
                mock_delete.assert_called_with(Bucket=..., Key=..., VersionId=...)  # tests most recent call
                mock_delete.reset_mock()
                assert gu.delete_glaciered_object_versions('bucket', 'key', delete_all_versions=True)
                assert mock_delete.call_count == 2
                mock_delete.assert_has_calls([
                    mock.call(Bucket=..., Key=..., VersionId=...),
                    mock.call(Bucket=..., Key=..., VersionId=...)
                ],
                any_order=True)
```
The `any_order=True` may not be necessary as you've written it but could be if you add parallelism.

dcicutils/glacier_utils.py
```diff
@@ -0,0 +1,455 @@
...
+        if parallel:
+            with ThreadPoolExecutor(max_workers=num_threads) as executor:
+                futures = []
+                for atid in atid_list:
+                    if isinstance(atid, dict):
+                        _atid = atid['@id']
+                        files_meta = self.resolve_bucket_key_from_portal(_atid, atid)
+                    else:
+                        _atid = atid
+                        files_meta = self.resolve_bucket_key_from_portal(_atid)
```
drio18, line 302: This general pattern of identifying the `@id` is repeated within this method and elsewhere and probably worthy of it's own helper method. May be worth it to reconsider the strategy of where the resolution of the string identifier vs. properties occurs such that it can be centralized in one place or as few locations as possible so separate method calls are not required (as is occurring with `resolve_bucket_key_from_portal` here). Consider renaming `atid_list` as well to reflect the fact that the contents are either `@id`s or properties.

dcicutils/glacier_utils.py
```diff
@@ -0,0 +1,455 @@
...
+                if phase == 1:
+                    current_success, current_failed = self.restore_portal_from_glacier(atid, item_meta,
+                                                                                       days=restore_length)
+                    if current_failed:
+                        failed.append(atid)
+                    else:
+                        success.append(atid)
+                elif phase == 2:
+                    raise GlacierRestoreException(f'Invalid phase for search_generator=True!'
+                                                  f' Do not use a generator when doing the copy phase.')
```
drio18, line 421: Is this only a bad idea when `parallel` is true? Seems like if not running in parallel there's no big difference between passing in the individual search results vs. all of them together.

dcicutils/glacier_utils.py
```diff
@@ -0,0 +1,455 @@
...
+            raise GlacierRestoreException(f'@id {atid} does not have an upload_key, thus cannot be queried for'
+                                          f' Glacier restore.')
+        # Add extra files
+        if 'extra_files' in atid_meta:
+            for extra_file in atid_meta['extra_files']:
+                if 'upload_key' in extra_file:
+                    files.append((bucket, extra_file['upload_key']))
+        return files
+
+    def restore_portal_from_glacier(self, atid: str, file_meta: Union[None, dict] = None,
```
drio18, line 109: Consider renaming this method for clarity, perhaps something like `get_portal_file_and_restore_from_glacier`.

dcicutils/glacier_utils.py
```diff
@@ -0,0 +1,459 @@
...
+                                              version_id: Union[str, None] = None) -> Union[dict, None]:
+        """ Reads the temporary location from the restored object and copies it back to the original location
+
+        :param bucket: bucket where object is stored
+        :param key: key within bucket where object is stored
+        :param storage_class: new storage class for this object
+        :param version_id: version of object, if applicable
+        :return: boolean whether the copy was successful
+        """
+        try:
```
drio18, line 232: Consider adding a check here to ensure the object to copy is one that has been restored from glacier (if possible with metadata on the object), otherwise there's no reason to perform such a copy and accumulate costs.

dcicutils/glacier_utils.py
```diff
@@ -0,0 +1,459 @@
...
+
+    def restore_glacier_phase_three_patch(self, atid_list: List[Union[str, dict]],
+                                          status: str = 'uploaded') -> (List[str], List[str]):
+        """ Patches out lifecycle information for @ids we've transferred back to standard
+
+        :param atid_list: list of @ids or actual file metadata objects to patch info on
+        :param status: top level status to replace for files
+        :return: 2 tuple of success, error list of @ids
+        """
+        success, errors = [], []
```
drio18, line 337: For phase 3 here and phase 4 below, consider double-checking a copied, non-glacier form of the object exists at the location prior to execution.

dcicutils/glacier_utils.py
```diff
@@ -0,0 +1,459 @@
...
+class GlacierRestoreException(Exception):
+    pass
+
+
+class GlacierUtils:
+
+    def __init__(self, env_name: str):
+        """ Pass an env_name that exists in your ~/.cgap_keys.json or ~/.fourfront_keys.json file
+            No support for admin keys!
+
```
drio18, line 29: It's hard to know without having used these methods yet, but naively I would think some of these methods could be abused, likely unintentionally, to cause issues or unintended effects by an inexperienced user that would result in unnecessary costs/incorrect metadata. I recommend some negative feedback that may be possible in further comments below, but I don't know if they are all possible with the information AWS provides. 

I don't think these are must-haves for now, but consider adding a warning line somewhere to this effect so users can know the lack of guardrails as well as maybe adding some TODOs for possible additions that could prevent unwanted actions.

dcicutils/glacier_utils.py
```diff
@@ -0,0 +1,459 @@
...
+        return success, fail
+
+    def restore_s3_from_glacier(self, bucket: str, key: str, days: int = 7) -> Union[dict, None]:
+        """ Restores a file from glacier given the bucket, key and duration of restore
+
+        :param bucket: bucket where the file is stored
+        :param key: key under which the file is stored
+        :param days: number of days to store in the temporary location
+        :return: response if successful or None
+        """
```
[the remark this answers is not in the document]
willronchetti: boto3 throws an error in that case, calling multiple restores while one is in progress appears harmless. After the restore has occurred though issuing another has the effect of extending the restore length.

dcicutils/glacier_utils.py
willronchetti: This isn't actually a copy, it's more like an overwrite (to change storage class), so no worries about increased cost (aside from those associated with the more expensive tier). It will fail if in an access tier that doesn't allow download I believe (ie: no restore active and in glacier or DA). This function is effectively just changing the storage tier.

dcicutils/glacier_utils.py
willronchetti: Agreed on "guard rails" generally... Will think on this more

dcicutils/glacier_utils.py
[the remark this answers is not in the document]
netsettler: I think the mock handles this case. That's what the .available_after in the mock is about.

PR #251, opened 2023-04-17
C4 1017 check certificate and access key

dmichaels-harvard:


dcicutils/scripts/publish_to_pypi.py
```diff
@@ -0,0 +1,302 @@
+# Script to publish the Python package in the CURRENT git repo to PyPi.
+# Does the following checks before allowing a publish:
+#
+# 1. The git repo MUST NOT contain unstaged changes.
+# 2. The git repo MUST NOT contain staged but uncommitted changes.
+# 3. The git repo MUST NOT contain committed but unpushed changes.
+# 4. The git repo package directories MUST NOT contain untracked files,
+#    OR if they do contain untracked files then you must confirm this is OK.
```
willronchetti, line 8: Very nice, love this

dcicutils/scripts/publish_to_pypi.py
```diff
@@ -0,0 +1,302 @@
...
+        return False
+    return True
+
+
+def verify_git_repo() -> bool:
+    """
+    If this (the current directory) looks like a git repo then return True,
+    otherwise prints a warning and returns False.
+    """
+    _, status = execute_command("git rev-parse --is-inside-work-tree")
```
willronchetti, line 117: I won't harp on this too much but there is a GitPython library that may allow you to these calls internally within Python (instead of subprocess which I assume you're doing internally).

dcicutils/scripts/publish_to_pypi.py
```diff
@@ -0,0 +1,302 @@
...
+    return False
+
+
+def exit_with_no_action() -> None:
+    """
+    Exits this process immediately with status 1;
+    first prints a message saying no action was taken.
+    """
+    PRINT("Exiting without taking action.")
+    exit(1)
```
willronchetti, line 298: This is used in other places no? May belong in a file in the actual library...

test/test_function_cache_decorator.py
```diff
@@ -0,0 +1,207 @@
...
+    assert f(3) == 9   # hit
+    assert f.cache_info()["size"] == 3
+    assert f(4) == 16  # hit
+    assert f.cache_info()["size"] == 3
+    assert f(5) == 25  # hit
+    assert f.cache_info()["size"] == 3
+    assert called == 3
+    assert f.cache_info()["hits"] == 6
+    assert f.cache_info()["misses"] == 3
+    assert f.cache_info()["size"] == 3
```
willronchetti, line 39: You might consider wrapping this in an object and making these properties so you don't need to index them like this

willronchetti (approved):
A few small comments, nothing major, looks great

dcicutils/scripts/publish_to_pypi.py
dmichaels-harvard: Hm, yeah, I sorta thought of looking into something like that but figured, at first anyways, to keep it as much like the existing bash script as possible, but maybe this would be better, may look into it, thanks.

test/test_function_cache_decorator.py
dmichaels-harvard: Yes, I like that - namedtuple - thanks.

dcicutils/scripts/publish_to_pypi.py
dmichaels-harvard: Maybe irrelevant point now, but before I realized (via Kent) that this could be shared, i.e. no need to be copied to each repo like the old publish bash script, my idea was that it should have absolute minimal dependencies, not even on dcicutils. But now that it is in dcicutils, I guess I could use dcicutils stuff, and including such a function as this that could go into misc_utils.

PR #254, opened 2023-05-10
Small Glacier Update

willronchetti:
- Updates `GlacierUtils` to remove `Lifecycle` tags from objects it is restoring (to prevent re-transfer by S3)

### DIFF
--- a/dcicutils/glacier_utils.py
+++ b/dcicutils/glacier_utils.py
@@ -2,7 +2,10 @@ import boto3
 from typing import Union, List, Tuple
 from concurrent.futures import ThreadPoolExecutor
 from tqdm import tqdm
-from .common import S3_GLACIER_CLASSES, S3StorageClass, MAX_MULTIPART_CHUNKS, MAX_STANDARD_COPY_SIZE
+from .common import (
+    S3_GLACIER_CLASSES, S3StorageClass, MAX_MULTIPART_CHUNKS, MAX_STANDARD_COPY_SIZE,
+    ENCODED_LIFECYCLE_TAG_KEY
+)
 from .command_utils import require_confirmation
 from .misc_utils import PRINT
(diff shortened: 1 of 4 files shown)


netsettler (approved):
I have more testing changes to add to this later, but I think this is an incremental improvement over what's on master, so I think it's fine to merge this if you're wanting to.

dcicutils/glacier_utils.py
```diff
@@ -342,12 +363,18 @@ def copy_object_back_to_original_location(self, bucket: str, key: str, storage_c
             response = self.s3.head_object(Bucket=bucket, Key=key)
             size = response['ContentLength']
             multipart = (size >= MAX_STANDARD_COPY_SIZE)
+            if not preserve_lifecycle_tag:  # default: preserve tags except 'Lifecycle'
+                tags = self.s3.get_object_tagging(Bucket=bucket, Key=key).get('TagSet', [])
+                tags = [tag for tag in tags if tag['Key'] != ENCODED_LIFECYCLE_TAG_KEY]
+                tags = self._format_tags(tags)
```
drio18, line 370: For files for which the only existing tag is the lifecycle tag, this results in `tags = ""`, which unfortunately then does not clear the tag. If the truthiness is `False`, something like this should work instead:
```suggestion
                tags = self.s3.get_object_tagging(Bucket=bucket, Key=key).get('TagSet', [])
                tags = [tag for tag in tags if tag['Key'] != ENCODED_LIFECYCLE_TAG_KEY]
                tags = self._format_tags(tags)
                if not tags:
                    self.s3.delete_object_tagging(Bucket=bucket, Key=key)
```
I used the above for some existing files, and the tags were cleared appropriately.

drio18 (changes requested):
Tags aren't cleared as is, but see comment for a solution that worked for me.

PR #264, opened 2023-06-07
Change to publish_to_pypi script to not allow publish using non-API-t…

dmichaels-harvard:
…oken username.


netsettler (approved):
Yeah, this looks fine. Don't forget to de-beta before merging.

PR #265, opened 2023-06-23
Add a license_utils module to dcicutils (C4-1048)

netsettler:
In service of [Programmatic checking of license compatibility (C4-1048)](https://hms-dbmi.atlassian.net/browse/C4-1048):

This adds a `license_utils` module to `dcicutils` that in particular will allow us to do tests like this in other repos:
 
```
from dcicutils.license_utils import C4InfrastructureLicenseChecker

def test_license_compatibility():
    C4InfrastructureLicenseChecker.validate()
```

NOTE WELL: This relies on the `pip-licenses` being a dev dependency of any repo wanting to do that. I don't know how to export a dev dependency, and am not sure you can, so there is a little dance at the top of the `license_utils.py` file that checks to see if the repo managed to be there at all, since we don't require it in the regular dependencies. The error message explains what needs to be done.

We now have a spreadsheet tracking the utility of this because it got complex.

I have extended this to have a `contribution_utils` as well, tracking contributors so that we can review them to make sure we know who's been involved.


dcicutils/license_utils.py
```diff
@@ -0,0 +1,806 @@
...
+    This set of values is useful to us in Park Lab where these tools were developed.
+    If you're at some other organization, we recommend you make a class that has values
+    suitable to your own organizational needs.
+    """
+
+    COPYRIGHT_OWNER = "President and Fellows of Harvard College"
+    LICENSE_TITLE = "The MIT License"
+    LICENSE_FRAMEWORKS = ['python', 'javascript']
+
+    ALLOWED = [
```
willronchetti, line 520: For our purposes it may be useful to be able to read this from file, so we can write top level files with this information or even a directory that can have sane defaults in utils and be overridden in other repos

test/test_license_utils.py
```diff
@@ -0,0 +1,691 @@
...
+
+        dummy_framework = LicenseFrameworkRegistry.LICENSE_FRAMEWORKS.get('dummy')
+        assert issubclass(dummy_framework, LicenseFramework)
+        assert dummy_framework is DummyLicenseFramework
+
+    finally:
+
+        # Clean up the mess we made...
+        for name in ['bogus_dummy', 'dummy']:
+            LicenseFrameworkRegistry.LICENSE_FRAMEWORKS.pop(name, None)
```
willronchetti, line 153: Better to use a function scope fixture for the registry no?

willronchetti (approved):
Might be smart to import/test in another repo before merging but generally looks great and super useful for sanity checking this info

dmichaels-harvard (approved):

test/test_license_utils.py
netsettler: Me? Ha. I never use fixtures except when in a tool that already does, to not fight the style. I hate their scoping rules. Usually I'd reach for a context manager instead because I think their scoping rules are clearer. Or I'd use a class with methods that have natural static scope.  I've rewritten this as a context manager in changes to be uploaded soon.

dcicutils/license_utils.py
netsettler: As a matter of design, I did not want individual repos to decide to override it. We can have several named policies. But the policies need to be approved organizationally, not at the level of coding lines.  I will likely add one for pipelines.

I have tried this in a number of repos (see the list on the first page). Some problems turned up that were since fixed. Others are ongoing issues for which I may create a way to mark the license as a known problem. We can talk more about this interactively perhaps sometime.

PR #266, opened 2023-06-28
S3 localstack ize

dmichaels-harvard:


dcicutils/boto_monkey_patching.py
```diff
@@ -0,0 +1,58 @@
...
+import boto3
+import os
+from typing import Optional
+
+
+_boto_client_original = boto3.client
+_boto_resource_original = boto3.resource
+_boto_service_overrides_supported = [
+    {"service": "s3", "env": "LOCALSTACK_S3_URL"},
+    {"service": "sqs", "env": "LOCALSTACK_SQS_URL"}
```
willronchetti, line 17: You might want constant for `LOCALSTACK_S3_URL` etc so can be imported and patched from pytest easily

test/conftest.py
```diff
@@ -1,4 +1,11 @@
 import os
+
+# Disable any AWS endpoint-url environment variables overrides for testing.
+if "S3_URL" in os.environ:
+    os.environ.pop("S3_URL")
+if "SQS_URL" in os.environ:
+    os.environ.pop("SQS_URL")
```
willronchetti, line 7: Does this not globally eliminate these values? You probably want a session scoped fixture that pops these values back on and off, just in case? Though probably not a big deal in local environment.

willronchetti (commented):
Code looks good but I think requires build/install instructions for localstack, will approve when that is added

dcicutils/boto_monkey_patching.py
dmichaels-harvard: Ah yes, thanks.

test/conftest.py
dmichaels-harvard: Yeah, I was a little unsure of this - was doing this because when I had this env vars set in my env, because I was using this facility for testing locally with localstack, tests were failing (because tests not setup of course with the assumption of using localstack), so I just globally unset these variables for all testing.

willronchetti (approved):

dcicutils/license_utils.py
```diff
@@ -853,6 +862,10 @@ class C4InfrastructureLicenseChecker(LicenseChecker):
             'typed-function',  # LICENSE at https://www.npmjs.com/package/typed-function?activeTab=code
 
         ],
+
+        'Other/Proprietary License': [
+            'localstack-ext'
```
netsettler, line 867: Needs a comment. Please don't add _any_ item to these lists without supporting rationale.

Also, needs a trailing comma so you don't have to change an extra line to add to the list.
```suggestion
            # This is known to be offered under Apache-2.0 license.
            # Ref: https://github.com/localstack/localstack/blob/master/LICENSE.txt
            'localstack-ext',
```

PR #268, opened 2023-07-05
EnvUtils SMaHT Support

willronchetti:
- Allow `smaht-portal` loading in EnvUtils

### DIFF
--- a/dcicutils/common.py
+++ b/dcicutils/common.py
@@ -10,6 +10,7 @@ REGION = 'us-east-1'
 
 APP_CGAP = 'cgap'
 APP_FOURFRONT = 'fourfront'
+APP_SMAHT = 'smaht'
 
 LEGACY_GLOBAL_ENV_BUCKET = 'foursight-test-envs'
 LEGACY_CGAP_GLOBAL_ENV_BUCKET = 'foursight-cgap-envs'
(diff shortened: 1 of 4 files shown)


dcicutils/common.py
```diff
@@ -10,6 +10,7 @@
 
 APP_CGAP = 'cgap'
 APP_FOURFRONT = 'fourfront'
+APP_SMAHT = 'smaht'
```
netsettler, line 13: You need to update lines 25 and 35 as well. See comment on lines 31-32. (I'm unable to attach a comment there.) It might be good to add a comment here that says
```
# If this set is updated, the values of ORCHESTRATED_APPS and OrchestratedApp need to be updated.
```

dcicutils/env_utils.py
```diff
@@ -425,11 +426,13 @@ def fresh_cgap_deployed_state_for_testing(cls):
             yield
 
     @classmethod
-    def app_case(self, *, if_cgap, if_fourfront):
+    def app_case(self, *, if_cgap, if_fourfront, if_smaht=None):
         if EnvUtils.ORCHESTRATED_APP == APP_FOURFRONT:
             return if_fourfront
         elif EnvUtils.ORCHESTRATED_APP == APP_CGAP:
             return if_cgap
+        elif EnvUtils.ORCHESTRATED_APP == APP_SMAHT:  # = None to avoid incompatibility
```
netsettler, line 434: You might alternatively want to initialize it before the `if` by doing:
```
if if_smaht is None:
    if_smaht = if_cgap
```
on the assumption that the cgap-like code is the best thing to use?

It might be prudent to make a similar change to `_orchestrated_app_case` in `env_utils_legacy.py`. It might not be reachable code, but it probably can't hurt.

netsettler (approved):
Please make the indicated changes before merging.

PR #269, opened 2023-07-17
Item model utils

drio18:


test/test_item_model_utils.py
```diff
@@ -0,0 +1,388 @@
...
+    def test_from_properties_and_existing_item(self) -> None:
+        properties = OTHER_ITEM_PROPERTIES
+        portal_item = get_portal_item()
+        result = PortalItem.from_properties_and_existing_item(properties, portal_item)
+        assert isinstance(result, PortalItem)
+        assert result.get_auth() == portal_item.get_auth()
+        assert result.should_fetch_links() == portal_item.should_fetch_links()
+
+
+def get_nested_property(
```
netsettler, line 331: I assume this is just making a sample property. I find this kind of thing hard to read because in test files, it's often hard to tell what the newly-defined functions to be tested are and what are the testing artifacts. I like to name things like this something like `make_sample_tested_property` so that in context of use, it's clear it's a testing artifact. Also, and it's a subtle point, but I like verbs like `get`, `find`, `lookup`, and `make` to have consistent meanings, and while we don't have a global convention for some of the subtleties I use for myself, I do think `get` usually connotes finding something that exists already, and here you're making something anew, which is why I'd use `make` as the part of the function name that's the verb, though maybe you're thinking of this as a representation of something that exists in the data? If so, I would leave it as `get` but I would clarify that in a doc string, which should be present in any case since this function does something non-trivial. 

Note that I'm not perfect at writing doc strings myself, but Will usually beats me up about it, presumably because it's hard to look at someone else's code and figure out what they're doing and he's usually on the outside of what I'm doing trying to claw his way in, as I'm doing here. :)

netsettler (commented):
A lot of these comments are stylistic. I apologize in advance for that, butI found this code a bit hard to read because of the chosen coding style, which doesn't match our normal style in other modules. I've tried to cite rationales for why we use the style we use. I fought with PEP8 for a long time when I came to Python but mostly feel like trying to normalize style makes it easier to both read and especially to grep.

There are a few non-syntactic comments in here, and I may have more later. I'd like to talk to you interactively about some of this stuff before taking another look.

test/test_item_model_utils.py
willronchetti: On the subject of docstrings, I really think some of these tests could use them and even some inline comments. They are quite complex relatively speaking and merit such documentation.

willronchetti (commented):
Kent left some good feedback that should be considered, not sure I have much more but I do really think it's worth documenting some of what's going on here especially in the more complex unit tests

PR #272, opened 2023-08-04
VariantUtils and TestVariantUtils

TomDuraisingh:
Functions that help filter down data from CGAP.


dcicutils/variant_utils.py
```diff
@@ -0,0 +1,88 @@
...
+    SEARCH_VARIANTS_BY_GENE = '/search/?type=VariantSample&limit=1&variant.genes.genes_most_severe_gene.display_title='
+    SEARCH_RARE_VARIANTS_BY_GENE = '/search/?samplegeno.samplegeno_role=proband&type=VariantSample\
+        &variant.csq_gnomadg_af_popmax.from=0&variant.csq_gnomadg_af_popmax.to=0.001\
+            &variant.genes.genes_most_severe_gene.display_title='
+
+    def __init__(self, *, env_name) -> None:
+        self._key_manager = CGAPKeyManager()
+        self.creds = self._key_manager.get_keydict_for_env(env=env_name)
+        # Uncomment this if needed
+        # self.health = get_health_page(key=self.creds)
```
willronchetti, line 17: I think good to remove at this point

dcicutils/variant_utils.py
```diff
@@ -0,0 +1,88 @@
...
+from dcicutils.ff_utils import get_metadata, search_metadata
+from dcicutils.creds_utils import CGAPKeyManager
+
+
+class VariantUtils:
+
+    SEARCH_VARIANTS_BY_GENE = '/search/?type=VariantSample&limit=1&variant.genes.genes_most_severe_gene.display_title='
+    SEARCH_RARE_VARIANTS_BY_GENE = '/search/?samplegeno.samplegeno_role=proband&type=VariantSample\
+        &variant.csq_gnomadg_af_popmax.from=0&variant.csq_gnomadg_af_popmax.to=0.001\
+            &variant.genes.genes_most_severe_gene.display_title='
```
willronchetti, line 11: I would prefer f-string for multi-line strings ie:
```
SEARCH_RARE_VARIANTS_BY_GENE = (f'/search/?samplegeno.samplegeno_role=proband&type=VariantSample'
                                f'&variant.csq_gnomadg_af_popmax.from=0&'
                                f'variant.csq_gnomadg_af_popmax.to=0.001&'
                                f'variant.genes.genes_most_severe_gene.display_title=')
```

dcicutils/variant_utils.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def get_total_result_count_from_search(self, gene):
+        """returns total number of variants associated with specified gene"""
+        res = get_metadata(self.SEARCH_VARIANTS_BY_GENE + gene, key=self.creds)
+        return res['total']
+
+    @staticmethod
+    def sort_dict_in_descending_order(unsorted_dict):
+        """sorts dictionary in descending value order"""
+        sorted_list = sorted(unsorted_dict.items(), key=lambda x: x[1], reverse=True)
```
willronchetti, line 41: sort dict how? answer: by value (key=lambda x: x[1]), so better name is `sort_dict_by_value_descending`, or better yet make the descending fact a default argument that can be changed so you can just have `sort_dict_by_value`. You can also make this one line by calling `dict` directly on line 41.

dcicutils/variant_utils.py
```diff
@@ -0,0 +1,88 @@
...
+        return dict(sorted_list)
+
+    def create_dict_of_mutations(self, gene):
+        """cretes dictionary of specified gene and 10+ occuring positions with their number of variants"""
```
willronchetti, line 45: Whenever you're creating a dict like this with specific structure, add info on the docstring on what the return value actually looks like.

dcicutils/variant_utils.py
```diff
@@ -0,0 +1,88 @@
...
+            pos = variant['variant']['POS']
+            if pos not in unique_positions:
+                unique_positions.add(pos)
+                mutation_dict[pos] = 1
+            else:
+                mutation_dict[pos] += 1
+        return {gene: self.sort_dict_in_descending_order({k: v for k, v in mutation_dict.items() if v >= 10})}
+
+    @staticmethod
+    def return_json(file_name):
```
willronchetti, line 58: Generally, it is useful to use type annotations ie:
```
def return_json(file_name: str) -> dict:
...
```
See [here](https://docs.python.org/3.9/library/typing.html) for more info. This applies to all functions in this file.

dcicutils/variant_utils.py
```diff
@@ -0,0 +1,88 @@
...
+        with open(file_name, 'r') as f:
+            file_content = json.loads(f)
+        return file_content
+
+    @staticmethod
+    def create_dict_from_json_file(file_name):
+        """creates dictionary object from json file"""
+        with open(file_name) as f:
+            json_list = f.read()
+        return json.loads(json_list)
```
willronchetti, line 68: Is there a practical need for this method along with the one above it?

dcicutils/variant_utils.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def create_list_of_msa_genes(self):
+        """creates list of all genes relating to the brain or nervous system (by 'neur' and 'nerv')"""
+        genes = self.return_json('gene.json')
+        return [gene['gene_symbol'] for gene in genes
+                if 'nerv' in gene.get('gene_summary', '')
+                or 'neur' in gene.get('gene_summary', '')]
+
+    def create_url(self, gene):
+        """returns a url to the variants at the most commonly mutated position of a gene"""
```
willronchetti, line 78: `create_url` isn't the best name I'd say... What you have in the docstring should probably be reflected in the name. As a general convention, for basic usage the function name should be as self explanatory as possible.

dcicutils/variant_utils.py
```diff
@@ -0,0 +1,88 @@
...
+        d = self.create_dict_from_json_file('10+sorted_msa_genes_and_mutations.json')
```
willronchetti, line 79: Hard-coded filename likely not desired

test/test_variant_utils.py
```diff
@@ -0,0 +1,134 @@
...
+        with self.mock_key_manager():
+            vu = VariantUtils(env_name='cgap_dummy')
+            mock_gene = 'GENE'
+            mock_create_dict_from_json_file.return_value = {
+                'GENE': {pos: 20, '123456': 10},
+                'OTHER_GENE': {pos: 10}
+            }
+            result = vu.create_url(gene=mock_gene)
+            expected_result = vu.SEARCH_RARE_VARIANTS_BY_GENE + mock_gene + ('&variant.POS.from={pos}\
+                                                                             &variant.POS.to={pos}&sort=-DP')
```
willronchetti, line 118: Likely to be much cleaner with f-strings ie:
```
expected_result = (f'{vu.SEARCH_RARE_VARIANTS_BY_GENE}{mock_gene}'
                   f'&variant.POS.from={pos}'
                   f'&variant.POS.to={pos}&sort=-DP')
```

It's also not clear this is working as desired, seems like you may want pos actually injected and that isn't happening as written I don't think?

test/test_variant_utils.py
```diff
@@ -0,0 +1,134 @@
...
+            assert result == expected_result
+            mock_create_dict_from_json_file.assert_called_once_with('10+sorted_msa_genes_and_mutations.json')
```
willronchetti, line 120: Hardcoded file path again - will break tests in the workflows

test/test_variant_utils.py
```diff
@@ -0,0 +1,134 @@
...
+
+    @patch('dcicutils.variant_utils.VariantUtils.return_json')
+    def test_create_list_of_als_park_genes(self, mock_return_json):
```
willronchetti, line 123: This tests looks very similar to another above... See if you can figure out how to parametrize them together as one

willronchetti (commented):
Releasing some feedback you can work on in background - feel free to reply/reach out with any questions or clarifications

willronchetti (approved):

PR #277, opened 2023-08-21
Support for getting a schema from a portal

netsettler:
* Add `misc_utils.to_camelcase`.
* Add `qa_utils.is_subdict`.
* Add `ff_utils.get_schema` and `ff_utils.get_schemas`.
* Add unit tests for bug [C4-1087](https://hms-dbmi.atlassian.net/browse/C4-1087) (marked `xfail`).

Other opportunistic changes:

* Minor tweaks to ``ff_utils.dump_results_to_json`` for style reasons, and repairs to its overly complex and error-prone unit test.
* Various tiny grammar changes comments and doc strings while I was in there in ``ff_utils`` because PyCharm grammar checker was fussing.

NOTES TO REVIEWERS:

* I am not sure why this endpoint is called profiles and not schemas, or what the difference is. So @willronchetti  had originally suggested some sort of `get_profile` function for `ff_utils` but I decided to call it `get_schema` instead since that better describes what I want. If there is a reason `get_profile` is a better name, that would be useful to know.
* Will had suggested a single function to get either a single schema or several, but that complicates the return type in the type hints, so I opted for separate functions to keep the simple function returning something that doesn't have to be peeled down further upon receipt. I don't really even need `get_schemas` so am open to a suggestion to just not include that, though I can see why it might be useful. It would be more useful if the profile operation took (maybe comma-separated?) several filenames.

[C4-1087]: https://hms-dbmi.atlassian.net/browse/C4-1087?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dcicutils/ff_utils.py
+++ b/dcicutils/ff_utils.py
@@ -6,7 +6,6 @@ import requests
 import time
 
 from collections import namedtuple
-from dcicutils.lang_utils import disjoined_list
 from elasticsearch.exceptions import AuthorizationException
 from typing import Optional, Dict, List
 from urllib.parse import parse_qs, urlencode, urlparse, urlunparse
@@ -16,7 +15,8 @@ from .common import (
     AnyAuthDict, AuthDict, SimpleAuthPair, AuthData, AnyAuthData, PortalEnvName,
     # S3BucketName, S3KeyName,
 )
-from .misc_utils import PRINT
+from .lang_utils import disjoined_list
+from .misc_utils import PRINT, to_camel_case
 
 
 # TODO (C4-92, C4-102): Probably to centralize this information in env_utils. Also figure out relation to CGAP.
@@ -281,7 +281,7 @@ def get_metadata(obj_id, key=None, ff_env=None, check_queue=False, add_on=''):
     Function to get metadata for a given obj_id (uuid or @id, most likely).
     Either takes a dictionary form authentication (MUST include 'server')
     or a string fourfront-environment.
-    Also a boolean 'check_queue', which if True
+    Also, a boolean 'check_queue', which, if True,
     will use information from the queues and/or datastore=database to
     ensure that the metadata is accurate.
     Takes an optional string add_on that should contain things like
@@ -421,7 +421,7 @@ def search_result_generator(page_generator):
     now return A,C,D but we already had the first page, so we request data starting at position 3
     for the second page and get E,G,I.  That means our sequence of return values would be A,C,E,E,G,I,K,M,
     or, in other words, showing a duplication. To avoid this, we keep track of the IDs we've seen
-    and show only the first case of each element, so A,C,E,G,I,K,M. (We won't see the D but we weren't
+    and show only the first case of each element, so A,C,E,G,I,K,M. (We won't see the D, but we weren't
     going to see it anyway, and it wasn't available the time we started, so the timing was already close.)
 
     Unfortunately, we aren't so lucky for deletion, though that happens more rarely. That will cause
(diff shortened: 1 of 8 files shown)


willronchetti (approved):
Looks useful, will look into the profiles issue on the server side as well as that behavior is not normal

drio18 (approved):

dcicutils/ff_utils.py
```diff
@@ -941,6 +942,40 @@ def _get_es_metadata(uuids, es_client, filters, sources, chunk_size, auth):
...
+    Returns:
+        dict: contains key schema names and value item class names
+    """
+    auth = get_authentication_with_server(key, ff_env)
+    url = f"profiles/{to_camel_case(name)}.json"
+    schema = get_metadata(url, key=auth, add_on='frame=raw')
+    return schema
+
+
 def get_schema_names(key=None, ff_env=None):
```
aschroed, line 979: Does it not make sense to call get_schemas in this function to remove redundant code?

aschroed (approved):
profiles is I think legacy or something maybe Burak and Alex B. came up with.  Don't know why but I don't see any problem with function names that you chose.

I have one non-blocking comment in line.

dcicutils/ff_utils.py
netsettler: Thanks for weighing in, @aschroed. It looks like `get_schema_names` does some extra filtering so isn't quite redundant, but I'm thinking _probably_ we don't need them to be disagreeing,. we just need to figure out which we want. My sense is that `get_schemas` should do that same filtering? That would make `get_schema_names` stay full compatible but with less code, as you suggest. (There's _some_ reason to ask why the endpoint doesn't do that filtering. but I'll count that as out-of-scope for now.) My intuition is that I should steal the additional filtering that `get_schema_names` does and have `get_schemas` only return the schemas that `get_schema_names` presently would name. Do you agree?

netsettler: @willronchetti and @drio18, feel free to venture an opinion.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "dcicutils"
-version = "7.8.0"
+version = "7.8.0.1b0"
```
dmichaels-harvard, line 3: Do you think it safe to update PyYML to ^6.0.1? I tried locally and tests pass, and I didn't see anything too scary in the release notes. It's just that everytime I build locally I have to manually update to this version due to Mac M1 issues with 5.4.1.

netsettler: Also, we can easily change to `">=5.1,<5.5,!=5.4.1"` if that would help you. Assuming it locks at all.

dcicutils/ff_utils.py
drio18: I would suggest not including the filtering in `get_schemas`, leaving it as is to have a broad, generic function; the retrieval of schemas plus the filtering could always be a separate function. I interpreted Andy's comment as refactoring `get_schema_names` to call  `get_schemas(key, ff_env)` rather than refactoring `get_schemas` to perform the filtering in `get_schema_names`.

aschroed: The extra filtering occurs after the redundant bit so I was just suggesting replacing 991-993 to a single call to get_schemas.  However, while investigating this I didn't find this function used at all and in fact I found a function of the same name that essentially does the same thing in the functions used in dcicwrangling notebooks - however, that function is now broken because of the fourfront schema upgrade.  And that function does not have any of the filtering. 
There is also a schema object in Submit4DN that does a similar thing but does not depend on the $id property so is not broken and not sure if it is worth changing that but the upshot is I need to do some refactoring of dcicwrangling and I will use these functions here and remove the redundant broken code there.   

I think it depends on the use case whether or not Abstract type schemas are included or not in the get_schemas function, I don't have a strong opinion but perhaps leaving filtering to calling functions makes most sense?

aschroed: Oh - and I have no idea about the local schemas without the id or $id property - maybe for testing only and should be updated to reflect actual schemas?

netsettler: I parameterized it. Can you look at what I did and see if it looks OK?

aschroed: Looks good to me.

PR #279, opened 2023-08-22
Python 3.11 upgrade

dmichaels-harvard:
Bump required `PyYAML` version to `^6.0.1` to resolve a Mac M1 issue with downstream builds.

As @netsettler notes in [discussion on utils PR #277](https://github.com/4dn-dcic/utils/pull/277#discussion_r1301683284), the [PyYAML Change Log](https://github.com/yaml/pyyaml/blob/master/CHANGES) says this about major version 6:

```
6.0 (2021-10-13)

* https://github.com/yaml/pyyaml/pull/327 -- Change README format to Markdown
* https://github.com/yaml/pyyaml/pull/483 -- Add a test for YAML 1.1 types
* https://github.com/yaml/pyyaml/pull/497 -- fix float resolver to ignore `.` and `._`
* https://github.com/yaml/pyyaml/pull/550 -- drop Python 2.7
* https://github.com/yaml/pyyaml/pull/553 -- Fix spelling of “hexadecimal”
* https://github.com/yaml/pyyaml/pull/556 -- fix representation of Enum subclasses
* https://github.com/yaml/pyyaml/pull/557 -- fix libyaml extension compiler warnings
* https://github.com/yaml/pyyaml/pull/560 -- fix ResourceWarning on leaked file descriptors
* https://github.com/yaml/pyyaml/pull/561 -- always require `Loader` arg to `yaml.load()`
* https://github.com/yaml/pyyaml/pull/564 -- remove remaining direct distutils usage
```

Based on this, he reasons:

> It looks like really dropping Python 2 is the major incompatible change, though the "always require Loader arg to yaml.load()" might require some research to make sure we weren't breaking something.


CHANGELOG.rst
```diff
@@ -7,6 +7,12 @@ Change Log
 ----------
 
 
+7.9.0
```
netsettler, line 10: I think a major version 8.0.0 would be wise, so that if there's an issue, it's easy for someone to use `^7.x.` kinds of notation to select stuff that doesn't have this until a fix is ready. Probably not an issue, but that thing about change in required arg worries me.

netsettler: @willronchetti do you have an opinion on this?

netsettler: I'm a little worried about that required arg. I note that if you [search for calls to yaml.load in 4dn-dcic](https://github.com/search?q=org%3A4dn-dcic%20yaml.load&type=code) (which is only one of our 3 orgs), you'll see several calls that don't pass an explicit Loader argument. Will that work or are you forcing code updates? Maybe those updates are fine. If there is an obvious loader, maybe we should have something that someone can call instead.

dmichaels-harvard: Yeah, looking at those results as well as for [dbmi-bgm](https://github.com/search?q=org%3Adbmi-bgm%20yaml.load&type=code) and [smaht-dac](https://github.com/search?q=org%3Asmaht-dac%20yaml.load&type=code), I don't see any uses within a repo where dcicutils is used (where the version of pyyaml it uses will increments due to this dcicutils change).

dmichaels-harvard: I built a cgap-portal(branch: dcicutils-with-pyyaml-6-and-sheet-utils | version: 14.2.0.1b1) with this dcicutils/pyyaml update and deployed to cgap-supertest and it seems fine (login/logout, create access-key, and ran submit-metadata-bundle).

pyproject.toml
```diff
@@ -36,7 +36,7 @@ classifiers = [
 
 
 [tool.poetry.dependencies]
-python = ">=3.7,<3.10"
+python = ">=3.7,<3.12"
```
netsettler, line 39:
```suggestion
python = ">=3.8,<3.12"
```

CHANGELOG.rst
```diff
@@ -6,6 +6,15 @@ dcicutils
...
 ----------
 
+8.0.0
+=====
+
+* Update Python to 3.11.
+* Changed PyYAML version to ^6.0.1; Mac M1 has issues building 5.4.1 (though 5.3.1 works).
+  See PyYAML 6.0 change log here: https://github.com/yaml/pyyaml/blob/master/CHANGES
+  The only incompatible change seems to be that yaml.load now requires a Loader argument;
+  and searching our repos (4dn-dcic, dbmi-bgm, smaht-dac) the only ones which might be affected
```
netsettler, line 16:
```suggestion
  and searching our GitHub organizations (4dn-dcic, dbmi-bgm, smaht-dac) the only ones which might be affected
```

netsettler (commented):
As @willronchetti noted on slack, testing workflows need to be updated to test the right things.

Also, the categories at top of pyproject.toml mention Python versions, so those need to change.

pyproject.toml
dmichaels-harvard: Done

CHANGELOG.rst
dmichaels-harvard: Done

netsettler (approved):
Yeah, this seems fine.

pyproject.toml
netsettler: This will also need update of the categories in pyproject.toml, and of the workflows in GA in order to get tests to pass (since 3.7 is now excluded) but also to test new versions we weren't previously testing.

pyproject.toml
```diff
@@ -70,7 +69,8 @@ coverage = ">=7.2.3"
 # coveralls = ">=3.3.1"
 flake8 = ">=3.9.2"
 flaky = ">=3.7.0"
-pip-licenses = "^3.5.5"
+pip-licenses = "^4.3.3"
+pip = "23.2.1"
```
netsettler, line 73: I would order this before `pip-licenses`, but that's minor.

Is it really a good idea to list `pip` here? Is `poetry` undoing the version we load in the `Makefile`? (In that case, maybe it is necessary to list it here. But it seems like something we may regret having to maintain.

Further, do you really mean to pin it to so specific a version? If so, could you add a comment saying why so we can know when to unpin it?

dmichaels-harvard: You're right, I had that (pip) in there temporarily - I will remove it. Thanks.

willronchetti (approved):
Approving recent changes for 3.11, all looks good but the pyproject python versions should be updated (take Kent's suggestion)

.github/workflows/main.yml
```diff
@@ -23,7 +23,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        python_version: [3.7, 3.8, 3.9]
+        python_version: [3.8, 3.9, 3.11]
```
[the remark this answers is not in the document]
dmichaels-harvard: Just to save time in GitHub actions really; not sure 3.10 is in wide usage; can add if you like.

PR #281, opened 2023-09-06
Allow ff_utils.get_schema and ff_utils.get_schemas to take a portal_vapp

netsettler:
First cut at implementing new args for `ff_utils.get_schema` and `ff_utils.get_schemas`: `portal_env=` and `portal_vapp=`.

Also opportunistically fixes a test failure related to an environment change that @willronchetti made.

### DIFF
--- a/dcicutils/ff_utils.py
+++ b/dcicutils/ff_utils.py
@@ -8,7 +8,7 @@ import time
 
 from collections import namedtuple
 from elasticsearch.exceptions import AuthorizationException
-from typing import Dict, List, Optional
+from typing import Optional, Dict, List
 from urllib.parse import parse_qs, urlencode, urlparse, urlunparse
 from . import s3_utils, es_utils
 from .common import (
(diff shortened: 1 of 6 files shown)


dcicutils/license_utils.py
netsettler, line 1: This change is not strictly needed for this PR, but it makes branch-switching easier for me so that test failures don't happen if I have a later venv.

test/test_s3_utils.py
netsettler, line 1: This is a fix for an environmental change made by @willronchetti.

dcicutils/ff_utils.py
netsettler, line 1: There are some small syntactic changes to doc strings in here that I made because the PyCharm grammar checker keeps pestering me about them. That's not really me being neurotic, just being annoyed that it is being so.

test/test_s3_utils.py
```diff
@@ -409,7 +409,7 @@ def test_s3utils_get_google_key():
     keys = s3u.get_google_key()
     assert isinstance(keys, dict)
     assert keys['type'] == 'service_account'
-    assert keys["project_id"] == "fourdn-fourfront"
+    assert keys["project_id"] == "fourfront-396315"
```
willronchetti, line 412: Should maybe add inline comment indicating this is a known constant

willronchetti (approved):

test/test_s3_utils.py
netsettler: I factored it out as a constant.

PR #282, opened 2023-09-07
Add portal_vapp= functionality to sheet_utils

netsettler:
This is a branch-to-branch PR that adds `portal_vapp=` functionality to the sheet_utils beta.

### DIFF
--- a/dcicutils/ff_utils.py
+++ b/dcicutils/ff_utils.py
@@ -8,7 +8,7 @@ import time
 
 from collections import namedtuple
 from elasticsearch.exceptions import AuthorizationException
-from typing import Optional, Dict, List
+from typing import Dict, List, Optional
 from urllib.parse import parse_qs, urlencode, urlparse, urlunparse
 from . import s3_utils, es_utils
 from .common import (
@@ -17,7 +17,7 @@ from .common import (
     # S3BucketName, S3KeyName,
 )
 from .lang_utils import disjoined_list
-from .misc_utils import PRINT, to_camel_case, remove_suffix
+from .misc_utils import PRINT, to_camel_case, remove_suffix, VirtualApp
 
 
 # TODO (C4-92, C4-102): Probably to centralize this information in env_utils. Also figure out relation to CGAP.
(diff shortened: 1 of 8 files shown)


dcicutils/ff_utils.py
```diff
@@ -942,41 +942,92 @@ def _get_es_metadata(uuids, es_client, filters, sources, chunk_size, auth):
...
+    portal_env = resolve_portal_env(ff_env=ff_env, portal_env=portal_env, portal_vapp=portal_vapp)
+    base_url = f"profiles/{to_camel_case(name)}.json"
+    add_on = 'frame=raw'
+    if portal_vapp:
+        full_url = f"{base_url}?{add_on}"
+        res = portal_vapp.get(full_url)
+        return get_response_json(res)
+    else:
+        auth = get_authentication_with_server(auth=key, ff_env=portal_env)
+        schema = get_metadata(obj_id=base_url, key=auth, add_on=add_on)
```
willronchetti, line 998: If you already have key, there is no need to call `get_authentication_with_server`?

dcicutils/ff_utils.py
```diff
@@ -942,41 +942,92 @@ def _get_es_metadata(uuids, es_client, filters, sources, chunk_size, auth):
...
-    schemas: Dict[str, Dict] = get_metadata('profiles/', key=auth, add_on='frame=raw')
+    portal_env = resolve_portal_env(ff_env=ff_env, portal_env=portal_env, portal_vapp=portal_vapp)
+    base_url = 'profiles/'
+    add_on = 'frame=raw'
+    if portal_vapp:
+        full_url = f"{base_url}?{add_on}"
+        schemas: Dict[str, Dict] = portal_vapp.get(full_url)
+    else:
+        auth = get_authentication_with_server(auth=key, ff_env=portal_env)
+        schemas: Dict[str, Dict] = get_metadata(obj_id=base_url, key=auth, add_on=add_on)
```
willronchetti, line 1030: Same comment here - duplicate call to `get_authentication_with_server` when `key` present

willronchetti (approved):
Approving but should review what I think are duplicate calls for authentication

dcicutils/ff_utils.py
netsettler: `get_authentication_with_server` only calls the server if you don't have the auth already. It's not a very good name. But also, this is how it worked before. I only added the other arm of this `if` statement.

netsettler: Fixed.

dcicutils/ff_utils.py
netsettler: Fixed.

PR #283, opened 2023-09-08
Fix to get_schema(s) to use URL prefixed with slash for vapp.

dmichaels-harvard:
- For the vapp (portal_vapp) case, the URL for ff_utils.get_schema(s) needs to be prefixed with a slash; maybe true in general but nervous about changing that now for fear of wider breakage.
- Also, the response from vapp.get returns a webtest.response.TestResponse, which has a json property which is an object rather than a function; handling this in ff_utils.get_response_json.

### DIFF
--- a/dcicutils/ff_utils.py
+++ b/dcicutils/ff_utils.py
@@ -990,7 +990,7 @@ def get_schema(name, key=None, ff_env: Optional[str] = None, portal_env: Optiona
     base_url = f"profiles/{to_camel_case(name)}.json"
     add_on = 'frame=raw'
     if portal_vapp:
-        full_url = f"{base_url}?{add_on}"
+        full_url = f"/{base_url}?{add_on}"
         res = portal_vapp.get(full_url)
         return get_response_json(res)
     else:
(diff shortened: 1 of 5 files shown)


dcicutils/ff_utils.py
```diff
@@ -1022,7 +1022,7 @@ def get_schemas(key=None, ff_env: Optional[str] = None, *, allow_abstract: bool
     base_url = 'profiles/'
     add_on = 'frame=raw'
     if portal_vapp:
-        full_url = f"{base_url}?{add_on}"
+        full_url = f"/{base_url}?{add_on}"
```
netsettler, line 1025: i guess this is most conservative. if `/` is ok in get_metadata, then maybe put the `/` in the initialization of `base_url`. But either is OK and this is safest, I guess.

dcicutils/ff_utils.py
```diff
@@ -1488,7 +1488,10 @@ def get_response_json(res):
     it is not present. Used with the metadata functions.
     """
     try:
-        res_json = res.json()
+        if isinstance(res, VirtualAppResponse):
+            res_json = res.json
+        else:
```
netsettler, line 1493: Again this is most conservative but I wonder if this should be type-checking, too. I guess let's go with this for now as most minimal change.

netsettler (approved):
You can ignore my whiny comments, which are not blocking but I just wanted to say. This is fine.

dcicutils/ff_utils.py
dmichaels-harvard: You mean type checking the `json` property?

netsettler: Oh, I was just wondering if we should do a plain else or another if with another type. But it previously just did `.json()` with no test, so this will be most compatible. I wouldn't have designed fresh code that way if I knew multiple types were in play. It preferences one default over another when you really don't konw. The presence of the `try:` around it is an apology for just taking a guess and hopes the damage won't be too great if it guesses wrong.

PR #284, opened 2023-09-14
License checker extensions

netsettler:
Some additions to license checkers.

* New license framework for R **(tentatively implemented)**
* New license framework for Conda **(tentatively implemented)**
* Much new license metadata
* New script `run-license-checker` for people to use this without installing it in code.

Having done this, though, we removed the `scan2-pipeline` policy definition because it can now be modularly added elsewhere, in that code. Such a definition can be added in an external module and then invoked interactively by:

```
run-license-checker scan2-pipeline --policy-dir <dir>
```

or from code by:

```
from dcicutils.license_utils import LicenseCheckerRegistry
Scan2PipelineLicenseChecker = LicenseCheckerRegistry.lookup_checker('scan2-pipeline', policy_dir=<policy-dir>)
Scan2PIpelineLicenseChecker.validate()
```

Note that if the conda checkers do not run, you may be in the wrong directory or Conda may not have set its `CONDA_PREFIX` environment variable. You can specify the directory by:
```
from dcicutils.license_utils import LicenseCheckerRegistry, LicenseOptions
Scan2PipelineLicenseChecker = LicenseCheckerRegistry.lookup_checker('scan2-pipeline', policy_dir="<policy-dir>")
with LicenseOptions.selected_options(conda_prefix="<conda-dir>"):
    Scan2PIpelineLicenseChecker.validate()
```
or you can override the search for Conda's `CONDA_PREFIX` environment variable by setting the `CONDA_LICENSE_CHECKER_PREFIX` environment variable, which can be used to override `CONDA_PREFIX`.


dcicutils/license_utils.py
```diff
@@ -117,25 +149,108 @@ def all_frameworks(cls):
...
+_PARENTHETICAL_VERSION_CONSTRAINT = re.compile(f'( [(]([>]?)(?:[=]|version) ([0-9.]+)({_OR_LATER_PATTERN}?)[)])')
+_POSTFIX_OR_LATER_PATTERN = re.compile(f"({_OR_LATER_PATTERN})")
+_GPL_VERSION_CHOICE = re.compile('^GPL-v?([0-9.+]) (?:OR|[|]) GPL-v?([0-9.+])$')
+
+
+def simplify_license_versions(licenses_spec: str, *, for_package_name) -> str:
+    m = _GPL_VERSION_CHOICE.match(licenses_spec)
+    if m:
+        version_a, version_b = m.groups()
+        return f"GPL-{version_a}-or-{version_b}"
```
willronchetti, line 166: This could use some docstring as the function name reads like its generic but you've got some GPL specific things going on here

dcicutils/license_utils.py
```diff
@@ -646,6 +954,109 @@ class C4InfrastructureLicenseChecker(LicenseChecker):
         'Zope Public License',
     ]
 
+    EXCEPTIONS = {
```
willronchetti, line 957: Should this be read from a file?

dcicutils/license_utils.py
```diff
@@ -757,141 +1168,242 @@ class C4InfrastructureLicenseChecker(LicenseChecker):
...
 class C4PythonInfrastructureLicenseChecker(C4InfrastructureLicenseChecker):
     """
-    For situations like dcicutils and dcicsnovault where there's no Javascript, this will test just Python.
+    Checker for C4 python library infrastructure (Fourfront, CGAP, SMaHT) from Park Lab.
     """
     LICENSE_FRAMEWORKS = ['python']
+
+
+@LicenseCheckerRegistry.register_checker('scan2-pipeline')
+class Scan2PipelineLicenseChecker(ParkLabGplPipelineLicenseChecker):
```
willronchetti, line 1376: Could you explain why this needs a separate class? Is this because it uses mixed software not living under one language?

willronchetti (approved):
These changes look good and the tool in general looks super useful and extendable as needed

dcicutils/license_utils.py
netsettler: It's true that it might not actually matter a lot if someone else used it. I can look at that.

dcicutils/license_utils.py
netsettler: I'm going to look into that. The trade-off is that these are really things that should be matters of fixed policy, at least insofar as the validator functions are, and people should not be thinking they can just casually edit this stuff to make it go away. It requires a lot of thought about legalities. But, that said, the command line form allows someone to explore whether they're compliant and to work toward getting things in order. For that it can be useful to evolve something and propose it as a coherent whole. So it's not a technical matter but a social, workflow, and legal issue. I'll do a little more functionality on that, though.

dcicutils/license_utils.py
netsettler: We only have this one worked case of someone new working on things, but my definite sense from the experience is that this is one of those places where exceptions are the most likely. Every now and then you can make exceptions on a broad basis, like for docutils, because it's simply the nature of docutils that people mostly use those at dev time. But for other libraries, the thing is that you only know you can make an exception if the use is conforming because of how that particular repo uses something, like we know we only use it in some maintenance script and it's not part of the core library that gets delivered to pypi, or something like that. So I can't put that kind of exception in the main classes. But there could be some other place these are kept. Maybe someplace maintained by an employer or a corporate division (in our case dbmi or hms).

dcicutils/license_utils.py
netsettler: This got done and is now merged to this PR.

PR #288, opened 2023-09-29
TransferUtils

willronchetti:
- A small addition to provision some baseline utilities for downloading files in parallel from the portal and from searches
- TODO: update with patching functionality once data model is in place

### DIFF
--- a/dcicutils/ff_utils.py
+++ b/dcicutils/ff_utils.py
@@ -299,6 +299,22 @@ def get_metadata(obj_id, key=None, ff_env=None, check_queue=False, add_on=''):
     return get_response_json(response)
 
 
+def get_download_url(obj_id: str, key: Optional[dict] = None, ff_env: str = None) -> str:
+    """
+    Function to get a download URL for a file without following the redirect so that such results can
+    be accumulated for retrieval later. Note that by default download URLs expire after 24 hours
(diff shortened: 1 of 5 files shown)


dcicutils/ff_utils.py
```diff
@@ -299,6 +299,22 @@ def get_metadata(obj_id, key=None, ff_env=None, check_queue=False, add_on=''):
...
+
+    :param obj_id: resource path to object we would like to download
+    :param key: relevant auth
+    :param ff_env: if using admin keys, name of env to resolve
+    :return: url to s3 for download
+    """
+    auth = get_authentication_with_server(key, ff_env)
+    get_url = '/'.join([auth['server'], _sls(obj_id)]).rstrip('/') + '/@@download'
+    response = authorized_request(get_url, auth=auth, verb='GET', allow_redirects=False)
+    return response.headers['Location']
```
dmichaels-harvard, line 315: Should this check for a 301?

willronchetti: Given the `@@download` at the end it will throw exception prior either way if it doesn't find it, so not sure there is a huge point (there is no scenario where 301 isn't returned and this code is reached, it would have thrown exception prior to this point)

PR #289, opened 2023-10-20
Better sheet utils

netsettler:
* New module ``bundle_utils.py`` that is intended for schema-respecting worksheets ("metadata bundle"). There are various modular bits of functionality here, but the main entry point here is:

  * ``load_items`` to load data from a given table set, doing certain notational canonicalizations, and checking that things are in the appropriate format.

* In ``common.py``, new hint types:

  * ``CsvReader``
  * ``JsonSchema``
  * ``Regexp``

* In ``lang_utils.py``:

  * New arguments ``just_are=`` to ``there_are`` get verb conjugation without the details.

  * Add "while" to "which" and "that" as clause handlers in the string pluralizer (e.g., so that "error while parsing x" pluralizes as "errors while parsing x")

* In ``misc_utils.py``, miscellaneous new functionality:

  * New class ``AbstractVirtualApp`` that is either an actual VirtualApp or can be used to make mocks if the thing being called expects an ``AbstractVirtualApp`` instead of a ``VirtualApp``.

  * New function ``to_snake_case`` that assumes its argument is either a CamelCase string or snake_case string, and returns the snake_case form.

  * New function ``is_uuid`` (migrated from Fourfront)

  * New function ``pad_to``

  * New class ``JsonLinesReader``

* In ``qa_checkers.py``:

  * Change the ``VERSION_IS_BETA_PATTERN`` to recognize alpha or beta patterns. Probably a rename would be better, but also incompatible. As far as I know, this is used only to not fuss if you haven't made a changelog entry for a beta (or now also alpha).

* New module ``sheet_utils.py`` for loading workbooks in a variety of formats, but without schema interpretation.

  A lot of this is implementation classes for each of the kinds of files, but the main entry point is intended to be ``load_table_set`` if you are not working with schemas. For schema-related support, see ``bundle_utils.py``.

* New module ``validation_utils.py`` with these facilities:

  * New class ``SchemaManager`` for managing a set of schemas so that programs asking for a schema by name only download one time and then use a cache. There are also facilities here for populating a dictionary with all schemas in a table set (the kind of thing returned by ``load_table_set`` in ``sheet_utils.py``) in order to pre-process it as a metadata bundle for checking purposes.

  * New functions:

    * ``validate_data_against_schemas`` to validate that table sets (workbooks, or the equivalent) have rows in each tab conforming to the schema for that tab.

    * ``summary_of_data_validation_errors`` to summarize the errors obtained from ``validate_data_against_schemas``.


dcicutils/glacier_utils.py
```diff
@@ -388,21 +381,16 @@ def copy_object_back_to_original_location(self, bucket: str, key: str, storage_c
...
-                    copy_args['CopySourceVersionId'] = version_id
+                    copy_target['CopySourceVersionId'] = version_id
                 if tags:
-                    copy_args['Tagging'] = tags
-                if self.kms_key_id:
-                    copy_args['ServerSideEncryption'] = 'aws:kms'
-                    copy_args['SSEKMSKeyId'] = self.kms_key_id
-                response = self.s3.copy_object(
-                    **copy_args, CopySource=copy_source
-                )
```
willronchetti, line 405: These deletions in glacier utils should be reverted I think, as KMS Key args are definitely needed. Not sure why these deletions are here.

test/data_files/sample_schemas/project.json
```diff
@@ -0,0 +1,592 @@
...
+        {
+            "$ref": "mixins.json#/status"
+        },
+        {
+            "$ref": "mixins.json#/static_embeds"
+        }
+    ],
+    "type": "object",
+    "properties": {
+        "static_headers": {
```
willronchetti, line 44: Same as with User, a lot of this schema stuff can probably be eliminated and just test representative cases.

test/test_bundle_utils.py
```diff
@@ -0,0 +1,682 @@
+import contextlib
+# import copy
+import glob
+import json
+import os
+import pytest
+import re
+
+# from collections import namedtuple
```
willronchetti, line 9: Some commented out imports you may want to clean up

dcicutils/sheet_utils.py
```diff
@@ -0,0 +1,679 @@
...
+                try:
+                    return float(value)
+                except Exception:
+                    pass
+        # If we couldn't parse it as an int or float, fall through to returning the original value
+        pass
+    return value
+
+
+def expand_string_escape_sequences(text: str) -> str:
```
willronchetti, line 101: Might be useful to docstring this since you're doing some fairly specific transformations

dcicutils/sheet_utils.py
```diff
@@ -0,0 +1,679 @@
...
+    def _all_rows(cls, sheet: Worksheet):
+        row_max = sheet.max_row
+        for row in range(2, row_max + 1):
+            yield row
+
+    @classmethod
+    def _all_cols(cls, sheet: Worksheet):
+        col_max = sheet.max_column
+        for col in range(1, col_max + 1):
+            yield col
```
willronchetti, line 359: For both of these are starting off the initial index - you should mention why and what the expected structure is. Functions with names like "all" here imply all to me, not necessary "all minus headers" or in the case of columns I'm not really sure what you're cutting off, all depends on expected structure.

willronchetti (commented):
Some small comments I think should be answered before approval, but generally looks great

test/test_bundle_utils.py
netsettler: I'll do that.

dcicutils/sheet_utils.py
netsettler: Yeah, more doc strings are needed. This is a good one to make sure I do.

dcicutils/sheet_utils.py
```diff
@@ -0,0 +1,679 @@
...
+        detected_encoding = bom_info and bom_info.get('encoding')  # tread lightly
+    use_encoding = 'utf-8' if detected_encoding == 'ascii' else detected_encoding
+    return io.open(filename, 'r', encoding=use_encoding)
+
+
+# TODO: Consider whether this might want to be an abstract base class. Some change might be needed.
+#
+# Doug thinks we might want (metaclass=ABCMeta) here to make this an abstract base class.
+# I am less certain but open to discussion. Among other things, as implemented now,
+# the __init__ method here needs to run and the documentation says that ABC's won't appear
```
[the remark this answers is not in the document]
netsettler: It also requires a certain kind of hygiene that I didn't really want to enforce because it's more theoretical than practical.

dcicutils/sheet_utils.py
netsettler: There is always a header in this format, so this is all of the data columns and rows. If there is not a header, none of this will work. I could rename this to `_all_data_rows` but it will not eliminate the risk, which is not checkable and must simply be documented. We don't have a submission protocol for non-headered files. TableSets, when I document them, must have headers, just as dictionaries cannot have keyless entries.

netsettler: Well, in the case of using this tool for items, we can check that the headings are things in the schema. I don't think it does that now, but it now could. data is unlikely to accidentally match. for tablesets, the lower level abstraction, there is no such reference so if you do:
```
1 2 3
4 5 6
7 8 9
```
you'll just get
```
[
    {'1': '4', '2': '5', '3': '6'},
    {'1': '7', '2': '8', '3': '9'}
]
```
and the effect won't be data loss, "just" data misuse. The first row isn't skipped, it's used as a header.

dcicutils/validation_utils.py
```diff
@@ -0,0 +1,290 @@
...
+        return re.findall(r"'(.*?)'", message)
+
+    schema_validator = jsonschema.Draft7Validator(schema)
+    for schema_validation_error in schema_validator.iter_errors(data_item):
+        if schema_validation_error.validator == "required":
+            errors.append({
+                "type": data_type,
+                "item" if identifying_value else "unidentified": identifying_value if identifying_value else True,
+                "index": data_item_index,
+                "missing_properties": schema_validation_error.validator_value})
```
dmichaels-harvard, line 225: I discovered this is actually not quite right; the list of missing properties can be gotten like this (I can make this change later if you want) ...` "missing_properties": list(set(schema_validation_error.validator_value) - set(schema_validation_error.instance))`

dcicutils/glacier_utils.py
dmichaels-harvard: Hm, not sure what this would have been about - I didn't make any changes related to glacier_utils.

dmichaels-harvard (approved):

PR #291, opened 2023-10-31
RAS Authentication Updates

utku-ozturk:
This PR extends the `value` that is stored as Redis `(key, value)` pair:

Old format: `jwtToken`
New format: `jwtToken:email` (since jwtToken is base64 encoded, :(colon as the separator makes sense. 

**ToDo**

From Researcher Auth Service (RAS) Project Partner Developer Guide:
`To comply with GA4GH, only the following signing algorithm is supported: RS256.`

While the data portals' Auth0 implementation works with HS256, RAS only uses RS256. Current JWT encoding/decoding functions throw the exception below since RS256 requires public/private keys to sign tokens. We bypassed the decoding's verify_signature by setting False for now, which should be True in production.

`ValueError: ('Could not deserialize key data. The data may be in an incorrect format, it may be encrypted with an unsupported algorithm, or it may be an unsupported key type (e.g. EC curves with explicit parameters).', [_OpenSSLErrorWithText(code=75497580, lib=9, reason=108, reason_text=b'error:0480006C:PEM routines::no start line')])`

**Related PRs:**

1. https://github.com/4dn-dcic/snovault/pull/273
2. https://github.com/4dn-dcic/fourfront/pull/1864
3. https://github.com/4dn-dcic/shared-portal-components/pull/238


dcicutils/redis_tools.py
```diff
@@ -93,27 +99,31 @@ def from_redis(cls, *, redis_handler: RedisBase, namespace: str, token: str):
...
         """ Decodes JWT to grab info such as the email
         :param audience: audience under which to decode, typically Auth0Client
         :param secret: secret to decrypt using, typically Auth0Secret
         :param leeway: numerical value in seconds to account for clock drift
         :return: a decoded JWT in dictionary format
         """
+        #TODO: verify_signature should be True
         return jwt.decode(self.jwt, secret, audience=audience, leeway=leeway,
-                          options={'verify_signature': True}, algorithms=['HS256'])
+                          options={'verify_signature': False}, algorithms=algorithms)
```
willronchetti, line 118: Does this still not work when `verify_signature` = True?

willronchetti (approved):
These changes generally look good but may merit some unit testing/fixes in existing tests, happy to take that on if needed. We also need to enable signature verification.

PR #293, opened 2023-11-20
SMaHT ingestion related work.

dmichaels-harvard:


dcicutils/zip_utils.py
```diff
@@ -0,0 +1,79 @@
...
+        yield tmp_directory_name
+
+
+@contextmanager
+def unpack_tar_file_to_temporary_directory(file: str) -> str:
+    with temporary_directory() as tmp_directory_name:
+        with tarfile.open(file, "r") as tarf:
+            tarf.extractall(tmp_directory_name)
+        yield tmp_directory_name
+
```
willronchetti, line 25: You might want to allow optional args to these functions in case additional options are needed/desired

dcicutils/zip_utils.py
```diff
@@ -0,0 +1,79 @@
...
+                    yield tmp_file_name
+
+
+@contextmanager
+def temporary_directory() -> str:
+    try:
+        with tempfile.TemporaryDirectory() as tmp_directory_name:
+            yield tmp_directory_name
+    finally:
+        remove_temporary_directory(tmp_directory_name)
```
willronchetti, line 57: I thought temp directories were removed automatically?

dcicutils/bundle_utils.py
```diff
@@ -55,38 +59,41 @@ def __repr__(self):
 
 
 class BoolHint(TypeHint):
+    def __init__(self):
+        super().__init__()
```
willronchetti, line 63: Is this not the default behavior?

dcicutils/bundle_utils.py
```diff
@@ -497,7 +562,15 @@ def contains_ref(self, item_type, item_ref):
...
         lookup_table = self.lookup_tables_by_tab_name.get(item_type)
         if lookup_table:  # Is it a type we're tracking?
-            return lookup_table.get(item_ref) or None
+            if isinstance(item_ref, list):
+                found = True
+                for item in item_ref:
+                    if not lookup_table.get(item):
+                        found = False
+                        break
+                return True if found else None
```
willronchetti, line 571: Why mixed return value (None + bool, why not just bool). Also, why not just 
```
for item in item_ref:
     if not lookup_table.get(item):
           return None
return True 
```

dcicutils/bundle_utils.py
```diff
@@ -523,16 +596,26 @@ def validate_ref(self, item_type, item_ref):
...
-            info = get_metadata(f"/{to_camel_case(item_type)}/{item_ref}",
-                                ff_env=self.portal_env, vapp=self.portal_vapp)
+            if isinstance(item_ref, list):
+                found = True
+                for item in item_ref:
+                    info = get_metadata(f"/{to_camel_case(item_type)}/{item}",
+                                        ff_env=self.portal_env, vapp=self.portal_vapp)
+                    if not isinstance(info, dict) or 'uuid' not in info:
+                        found = False
+                        break
```
willronchetti, line 606: Same pattern here as above can be simplified

willronchetti (approved):
Some small comments, nothing major

dcicutils/bundle_utils.py
dmichaels-harvard: No it's not. A derived __init__ must explicitly call its base __init__, if it wants that behavior.

dcicutils/zip_utils.py
dmichaels-harvard: Oh yeah, wait a minute, I think you should be right; I think I did this for a reason, need to remind myself.

PR #298, opened 2024-01-22
Changes to itemize SMaHT submission ingestion create/update/diff situation.

dmichaels-harvard:
Changes to itemize SMaHT submission ingestion create/update/diff situation;
and to handle property deletions in submission spreadsheet.


willronchetti (approved):
When you have time I'd highly recommend writing unit tests for the "compare" etc functions

PR #304, opened 2024-04-12
Minor fix in structured_data to not try to resolve empty refs.

dmichaels-harvard:
* Minor fix in structured_data to not try to resolve empty refs in norefs mode;
  and added StructuredDataSet.unchecked_refs; not functionally substantive as
  used (only) with smaht-submitr/submit-metadata-bundle --info --refs.
* Added nrows and nsheets to data_reader; convenience for smaht-submitr/submit-metadata-bundle --info.
* Added test_progress_bar module for progress_bar testing; would like to add more tests.
* Fixed up captured_output module to handle UTF-8 encoding to help unit testing progress_bar.
* Added hooks to progress_bar to help unit testing.
* Added a find_nth_from_end and set_nth to misc_utils to help progress_bar unit testing.
* Added format_size and format_duration misc_utils; refactor from smaht-submitr.
* Added format_datetime, parse_datetime to datetime_utils; refactor from smaht-submitr; and some tests.
* Added check_only flag to portal_utils.Portal.{post,patch}_metadata (came up in ad hoc troubleshooting).


dcicutils/captured_output.py
```diff
@@ -68,3 +68,19 @@ def uncaptured_output():
     finally:
         sys.stdout = original_stdout
         sys.stderr = original_stderr
+
+
+class _EncodedStringIO:
```
willronchetti, line 73: Might be worth documenting why this was done

dcicutils/structured_data.py
```diff
@@ -154,7 +154,32 @@ def resolved_refs(self) -> List[str]:
...
+        """
+        def load_json(value: str) -> Optional[dict]:
+            try:
+                return json.loads(value)
+            except Exception:
+                return None
+        result = []
+        if self._norefs:
+            for ref in self._resolved_refs:
+                if len(ref) >= 3 and (ref_path := ref[0]) and (ref_src := load_json(ref[2])):
```
willronchetti, line 177: Do you mind explaining this? I'm having trouble following this code

willronchetti (approved):
Looks good, some small questions

dcicutils/structured_data.py
dmichaels-harvard: Yeah this is should be refactored a bit. I've added this comment, hopefully clears it up a bit .. 
# The structure of this self._resolved_refs is setup in Schema._map_function_ref,
# which is called whenever a reference (linkTo) is encountered. It is a set of
# tuples containing three items: [0] the ref path, [1] its uuid (if applicable),
# and [2] its src. The src identifies the place where this ref occurred and is a
# dictionary containing file, type, column, and row properties. For this case, of
# unchecked (norefs) references, the uuid ([1]) is None because we are skipping
# ref resolution. And the src is actually a *string* dump of the dictionary, only
# because dictionarties cannot be put in a set (which is what _resolved_refs is);
# this dump is also done in Schema._map_function_ref (should probably change this
# to be a list to avoid this - TODO); we only even store this src info for this
# norefs case, as not really needed otherwise. This is just to support the
# smaht-submitr/submit-metadata-bundle --info --refs options.

PR #307, opened 2024-05-21
Fix key manager registry

drio18:
This PR changes the KeyManager classes added to the registry to be the portal-specific KeyManagers rather than the generic KeyManager which cannot be used.


willronchetti (approved):
Needs version bump/CHANGELOG

PR #312, opened 2024-07-08
Added numpy ^1.26.4 in pyproject.toml

dmichaels-harvard:


dcicutils/scripts/update_portal_object.py
```diff
@@ -0,0 +1,430 @@
...
+# Schema properties to ignore (by default) for the view schema usage.
+_SCHEMAS_IGNORE_PROPERTIES = [
+    "date_created",
+    "last_modified",
+    "principals_allowed",
+    "submitted_by",
+    "schema_version"
+]
+
+_SCHEMA_ORDER = [  # See: smaht-portal/src/encoded/project/loadxl.py
```
willronchetti, line 50: What do you use this for? I would recommend not doing it this way, instead add an API in smaht-portal that returns this and have this info come in via a call to the server

willronchetti (approved):

PR #318, opened 2024-10-04
Added optional bucket argument to s3_utils.read_s3 function.

dmichaels-harvard:
* Added optional bucket argument to s3_utils.s3Utils.read_s3 function (initially for Andy).


aschroed (approved):
You beat me to it - I was adding the same and happy to go with your version!

PR #331, opened 2026-05-05
SN Refactor custom_excel for portal query

sarahgonicholson:
- Update `custom_excel.py` to use a portal query for GenericQcConfig items rather than use the `config/custom_column_mappings.json` file 

Used Claude for help writing this, and it is difficult to test because it is only called within submitr


willronchetti:
Follow-up: hardened the setuptools dependency so a future `poetry update` cannot re-break the build.

**Root cause of the earlier CI failures** (already fixed by the re-lock commit): the branch's `poetry update` floated tooling unrelated to the boto bump — `setuptools` 70.3.0 → 83.0.0 (setuptools **82 removed `pkg_resources`**, which we import at module load in `deployment_utils.py`/`project_utils.py` and which pinned `pyramid 1.10.8` imports in `pyramid.asset` → 6 collection errors), and `flake8` 7.1.0 → 7.3.0 (pyflakes 3.2→3.4 enabled F824 on 43 pre-existing dead `nonlocal`/`global` declarations → failed the py3.12 flake8 step). The re-lock restored both to their known-good versions while keeping the intended boto3/botocore 1.43.58 bump.

**This commit adds the requested durable pin:**
- `setuptools = ">=70.3,<82"` in `[tool.poetry.dependencies]` (setuptools is a genuine runtime dep since shipped code imports `pkg_resources`). The lock still resolves 70.3.0; only the content-hash changes.
- `test/test_setuptools_version.py` — a regression guard asserting the resolved setuptools major stays `< 82`. It imports only `setuptools` (not `pkg_resources`), so a widened pin fails with a clear assertion rather than the opaque "No module named 'pkg_resources'" collection error. Verified executing + passing in both the 3.11 and 3.12 CI jobs.

**Please note (no CI job on this repo can detect it):** the `<82` ceiling propagates to downstream consumers (fourfront, cgap-portal, snovault, submitr) once a new dcicutils is released — any consumer already resolving setuptools ≥82 will fail to resolve dcicutils until its own `pkg_resources` uses are migrated. The 43 F824 dead-declaration sites are real and worth a separate cleanup; they are intentionally out of scope here. `python = ">=3.10,<3.13"` (dropping 3.8/3.9) was inherited from the boto-constraint commit, not introduced here.

All checks green: Static ✓, Python 3.11 ✓, Python 3.12 ✓.

willronchetti (approved):

